PR #10, opened 2021-11-12
Update pydata-sphinx-theme to allow update and fix crash in sphinx docs build

germa89:


setup.py
```diff
@@ -40,9 +40,7 @@ def read(rel_path):
     include_package_data=True,
     # See http://www.sphinx-doc.org/en/stable/theming.html#distribute-your-theme-as-a-python-package
     entry_points={"sphinx.html_themes": ["pyansys_sphinx_theme = pyansys_sphinx_theme"]},
-    install_requires=["sphinx",
-                      "pydata-sphinx-theme==0.6.3",
```
akaszynski, line 44: You still need the theme.

germa89: I added it to the requirements now. I realized later.

germa89:
@akaszynski I guess now.... ?

akaszynski:
No, because the theme install requires must have the pydata theme.  You should just upgrade the install requires.  See suggestions.

requirements_docs.txt
```diff
@@ -1 +1,2 @@
 sphinx==4.0.3
+pydata_sphinx_theme
```
akaszynski, line 2:
```suggestion
sphinx
```

akaszynski:
We no longer need install requirements_docs, right?

germa89:
True

akaszynski:
You need to change the workflow in .github/workflows

akaszynski (approved):
LGTM

setup.py
```diff
@@ -40,9 +40,7 @@ def read(rel_path):
...
-                      ],
+    install_requires=["sphinx"],
```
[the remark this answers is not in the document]
germa89: I guess better not fix the version ?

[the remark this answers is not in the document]
akaszynski: It's better to fix the version because if we don't incorporate upstream changes, we run the risk of breaking something.

PR #14, opened 2021-12-10
Improve API doc empty-space formatting

Andy-Grigg:
Adds some custom styling to improve the grouping and readability of generated API docs. Side by side comparison (current on left, proposed on right):

image

There's less space between 'Parameters' and 'Examples', and more space between 'Examples' and the methods. This helps the examples feel more like they belong to the class itself, as opposed to the method below.

I'm not a CSS expert, but I think the selectors are appropriate and seem to align with how the styling is written in the base sphinx package. These changes have been tested on the example API Documentation in this package, and on the as-yet-unreleased grantami-bomanalytics package.

### DIFF
--- a/pyansys_sphinx_theme/static/css/theme.css
+++ b/pyansys_sphinx_theme/static/css/theme.css
@@ -166,3 +166,16 @@
 i.fa-github-square:before {
     color: #ffffff !important;
 }
+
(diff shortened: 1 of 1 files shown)


akaszynski (approved):
Agree with the style changes.  Thanks for adding this. Ping me if you need this now of it can wait until the next release (either minor or patch).

PR #17, opened 2022-01-21
Added style to table representation

germa89:
Hi there

Since we would like to use Pandas DataFrame to present data in PyMAPDL documentation, I thought we might as well include the style in the Pyansys Theme.

The proposal is shown as this:

image

I'm happy to do any changes you might propose. I just tried the colors align with Ansys color theme.

### DIFF
--- a/pyansys_sphinx_theme/static/css/theme.css
+++ b/pyansys_sphinx_theme/static/css/theme.css
@@ -218,6 +218,56 @@ tr:nth-child(even){
 background-color: rgb(255, 183, 27, 0.15); 
 }
 
+/*
+###############
+   longtable 
+###############
+
+Use as:
+
+.. table:: 
+  :class: table-centered
+
+  +---------------------------+-------------------+
+  |                           | MAPDL Command     |
+  +===========================+===================+
+  | **GUI commands**          | * ``*ASK``        |
+  |                           +-------------------+
+  |   ...
+  |   ...
+
+
+*/
+table.table-centered {
+  width:100%;
+  max-width: 100%; 
+  border-spacing: 0;
+  border-collapse: collapse;
+  overflow: hidden;
+  vertical-align: middle;
+
+  /* Disabling scroll bars */
+  overflow-y: scroll;
+  scrollbar-width: none; /* Firefox */
+  -ms-overflow-style: none;  /* Internet Explorer 10+ */
+  }
+
+.table-centered tr { 
+  background-color: white;
+}
+
+.table-centered th {
+  background-color: rgb(255, 183, 27, 0.55);
+}
+
+.table-centered tr:nth-child(odd){ 
+  background-color: white;
+}
+
+.table-centered tr:nth-child(even){
+background-color: rgb(255, 183, 27, 0.15); 
+}
+
 
 /*
 ###############
@@ -230,26 +280,58 @@ table.longtable tr:hover td {
 }
 
 /*
-###############
-   DataFrame 
-###############
-Not sure why the odd columns (white) are rendered gray.
+#########################
+   longtable-centered
+#########################
+
+Use as:
+
+
+.. table:: 
+  :class: longtable-centered
+
+  +---------------------------+-------------------+
+  |                           | MAPDL Command     |
+  +===========================+===================+
+  | **GUI commands**          | * ``*ASK``        |
+  |                           +-------------------+
+  |   ...
+  |   ...
+
 */
 
-.dataframe tr{ 
-    background-color: white;
-  }
-
-.dataframe tr:nth-child(odd){ 
-    background-color: white;  
-  }
-
-.dataframe tr:nth-child(even){
-  background-color: rgb(255, 183, 27, 0.15);
+table.longtable-centered tr:hover td {
+  background-color: rgb(255, 183, 27, 0.6);
+  text-align: center;
 }
 
-.dataframe tr:hover td {
+/*
+#########################
+   DataFrame-centered
+#########################
+Not sure why the odd columns (white) are rendered gray.
+
+I do not think we will be able to call this class.
+*/
+
+.dataframe-centered tr{ 
+    background-color: white;
+    text-align: center;
+  }
+
+.dataframe-centered tr:nth-child(odd){ 
+    background-color: white;  
+    text-align: center;
+  }
+
+.dataframe-centered tr:nth-child(even){
+  background-color: rgb(255, 183, 27, 0.15);
+  text-align: center;
+}
+
+.dataframe-centered tr:hover td {
   background-color: rgb(255, 183, 27, 0.6);
+  text-align: center;
 }
 
 
@@ -270,5 +352,7 @@ Not sure why the odd columns (white) are rendered gray.
 .autosummary tr:nth-child(even){
   background-color: white;
 }
+
+
 
 
--- a/pyansys_sphinx_theme/static/css/theme.css
+++ b/pyansys_sphinx_theme/static/css/theme.css
@@ -179,34 +179,84 @@ p.rubric {
     margin-top: 0.75em;
     margin-bottom: 0.75em;
 }
+
+
 /*
-Table style for the DataFrame representation
+#############
 */
 table {
-  border-collapse: collapse;
-  width: 100%;
-}
+        width:100%;
+        max-width: 100%; 
+        border-spacing: 0;
+        border-collapse: collapse;
+        overflow: hidden;
+        z-index: 1;
+        font-size: 16px;
+        text-align: center;
+        
+        /* Disabling scroll bars */
+        overflow-y: scroll;
+        scrollbar-width: none; /* Firefox */
+        -ms-overflow-style: none;  /* Internet Explorer 10+ */
+        }
 
-th {
-  text-align: center;
-  padding: 8px;
-  font-size: 16px;
-}
 
-td {
-  text-align: left;
-  padding: 8px;
-  font-size: 16px;
-}
+ /* Disabling scroll bars */
+table::-webkit-scrollbar { /* WebKit */
+    width: 0;
+    height: 0;
+}    
 
-td:hover {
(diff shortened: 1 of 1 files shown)


germa89:
Slightly improved version:

image


## Notes
This affects any table in the documentation (see for example tech demo in ``User guide`` > ``Extended examples``). It looks super good in my opinion, specially for big tables. 
image

The issue is it **ALSO** affects _autosummary_ (not sure why).

image

This is not a good effect.

### Proposal
I wonder if someone more skilled than me can sort out how to create a table style in CSS which is only applied to Dataframes.
I believe by default the style for dataframes is something like ``Dataframe table`` or similar (please check this). If it is like that, we should just over write that style in our ``ansys.css``. 

I did a bit of research and I didn't find the way to do this. However, if requested I can keep trying.

pyansys_sphinx_theme/static/css/theme.css
```diff
@@ -179,3 +179,84 @@ p.rubric {
     margin-top: 0.75em;
     margin-bottom: 0.75em;
 }
+
+
+/*
+#############
+*/
+table {
```
Andy-Grigg, line 187: It looks like all dataframe tables have a `dataframe` class, so you should be able to do something like `table.dataframe {`, and have it just select dataframe tables.

pyansys_sphinx_theme/static/css/theme.css
```diff
@@ -179,3 +179,84 @@ p.rubric {
...
+
+
+ /* Disabling scroll bars */
+table::-webkit-scrollbar { /* WebKit */
+    width: 0;
+    height: 0;
+}    
+
+/* Headers and data cells format */
+    td, th {
```
Andy-Grigg, line 211: I think you'll need to do something like `table.dataframe > td, th` here, to get only td or th elements who have a parent `<table class="dataframe">` element.

pyansys_sphinx_theme/static/css/theme.css
```diff
@@ -179,3 +179,84 @@ p.rubric {
...
+        overflow: hidden;
+        z-index: 1;
+        font-size: 16px;
+        text-align: center;
+        
+        /* Disabling scroll bars */
+        overflow-y: scroll;
+        scrollbar-width: none; /* Firefox */
+        -ms-overflow-style: none;  /* Internet Explorer 10+ */
+        }
```
Andy-Grigg, line 201: This file could do with running through a formatter to clean up the whitespace

germa89: Hi @Andy-Grigg 

I'm not sure how to do that. Can you point me to some resource?

Kind regards,

pyansys_sphinx_theme/static/css/theme.css
germa89: I'm trying now

Andy-Grigg (commented):
These changes are nice! I like the colors and the highlighting.

I'm seeing some slight weirdness with building the docs for an as-yet-unreleased package I'm working on, I'll follow up with you separately on what I'm seeing there.

I am by no means a css expert, but I've added some comments that I think should help limit the changes to affect dataframes only.

pyansys_sphinx_theme/static/css/theme.css
germa89: No idea what you mean :-/

germa89: Oh.... I will no problem!

germa89:
# Final proposal:

- The general style (colouring) is common to tables, dataframes and longtable.
image

- Tables do not have hovers.
image

- Dataframes and longtable do have hovers.
image

- autosummary tables have all their rows white and do have hovers (the rest of the format is inherited from table)
image

pyansys_sphinx_theme/static/css/theme.css
```diff
@@ -179,3 +179,180 @@ p.rubric {
...
+
+/*
+###############
+   longtable 
+###############
+
+Use as:
+
+.. table:: 
+  :class: table-centered
```
Andy-Grigg, line 229: These comments are a good idea! It's not obvious otherwise how the ReST relates to the CSS

pyansys_sphinx_theme/static/css/theme.css
```diff
@@ -179,3 +179,180 @@ p.rubric {
...
+  |                           | MAPDL Command     |
+  +===========================+===================+
+  | **GUI commands**          | * ``*ASK``        |
+  |                           +-------------------+
+  |   ...
+  |   ...
+
+
+*/
+table.table-centered {
```
Andy-Grigg, line 241: This block looks the same as the `table` above, is it possible to merge them?

pyansys_sphinx_theme/static/css/theme.css
```diff
@@ -179,3 +179,180 @@ p.rubric {
...
+tr { 
+  background-color: white;
+}
+
+th {
+  background-color: rgb(255, 183, 27, 0.55);
+}
+
+tr:nth-child(odd){ 
+  background-color: white;
```
Andy-Grigg, line 214: I don't know if you are seeing this as well, but this declaration is getting overridden by something within the html itself. If that's the case, I think the only way to get around it is to add `!important`, something like:

`background-color: white !important;`

!important can cause problems, because it overrides the natural organization and structure of CSS. Sometimes though, there is no alternative...

Andy-Grigg:
I have added a few more comments. I like what you've done with the comments and adding in specific sections for different types of tables, it's a bit more verbose but I think it's the right way to go.

I have only tested the `dataframe` table type, because that's the only one our docs are currently using. I'll have to rely on you or Alex to test out some of the other table types you have added to the CSS.

akaszynski:
> I have only tested the dataframe table type, because that's the only one our docs are currently using. I'll have to rely on you or Alex to test out some of the other table types you have added to the CSS.

I typically use sphinx tables and `pandas.DataFrame`, so we're going to have to test out the other styles as we go.

I think these tables look excellent. Nice work @germa89 (and by extension, @RomanIlchenko1308 who I think helped inspire this work with VM examples).

pyansys_sphinx_theme/static/css/theme.css
germa89: mmhh it is interesting that for the ``_autosummary`` definition, the white works.

We could use important for the moment, until we see any trouble or find a better solution.

pyansys_sphinx_theme/static/css/theme.css
germa89: Thank you!

pyansys_sphinx_theme/static/css/theme.css
germa89: I forgot to add ``text-align: centered`` to this block.

pyansys_sphinx_theme/static/css/theme.css
germa89: I'm checking again... I think it is fine with the light gray (it is very light), so I think we can leave it as it is and move on. It shows a bit but if you believe we are going to have side effects, I'm fine with that gray around.

germa89: i'm not able to get the white color in the odd rows of the dataframe, even if i use ``!important``. I think we can leave it like that for the moment. 

I guess this is going to require review the whole CSS or something else which I have not much expertise.

germa89:
> I typically use sphinx tables and pandas.DataFrame, so we're going to have to test out the other styles as we go.

In PyMAPDL you can find:
- *Tables*: https://dev.mapdldocs.pyansys.com/user_guide/mapdl.html#unsupported-interactive-commands
- *Long tables*: https://dev.mapdldocs.pyansys.com/user_guide/mapdl.html#table-1
- *Plain Dataframes*: vm8 (when merged) https://mapdldocs.pyansys.com/examples/06-verif-manual/vm-008-parametric_calculation.html#sphx-glr-examples-06-verif-manual-vm-008-parametric-calculation-py
- and *Formatted Dataframes*: https://dev.mapdldocs.pyansys.com/examples/06-verif-manual/vm-007-plastic_compression_of_a_pipe_assembly.html#sphx-glr-examples-06-verif-manual-vm-007-plastic-compression-of-a-pipe-assembly-py

If this theme modification is successful I guess we will get rid of formatted dataframes, in favour of plain dataframes which will inherit the settings (colors, etc) from the theme.

pyansys_sphinx_theme/static/css/theme.css
Andy-Grigg: Yea, let's not go too far. The best CSS is the CSS that works!

RomanIlchenko1308:
Hi @germa89 and @Andy-Grigg,

I have checked your modification of table style, and I have seen this not an issue,
however, it looks weird a bit,
that we have a connectivity (hover selection) of the first row and the index column with column, as you can see below:

image

But, if you move you mouse down, we have just a partial hover selection of the columns but without indexing column as you can see it below:

image


If I am wrong, please write me)
thank you in advance @germa89 and @Andy-Grigg !!!

Andy-Grigg:
Hi @RomanIlchenko1308!

I think the issue is caused by the merged rows, and I'm not really sure what the best option is to fix it. The first column is taking the color of the row that it starts on. The same thing is happening when it highlights as well.

I don't think there's much we can do to make the first column highlight correctly, because the CSS can only ever highlight a cell all the same color. I think all we can do is _not_ highlight tables with merged cells, but I don't know if that's possible. @germa89, any thoughts?

germa89:
We could create a class for non-highlighted tables if needed, so we leave it at the writer if he wants that behavior or not.

Other than that, I do not think I can fill the color issue. It is way beyond my expertise. I could try, but the amount of time required vs reward, it is very low. 

I would leave it as it is. Or if @RomanIlchenko1308 wants to implement the non-highlighted class, I'm happy to review them.

PR #18, opened 2022-01-24
Rethinking code style

germa89:
I'm not super happy about the readability of the text rendered as code:

image

And some people (@RomanIlchenko1308 ) have already pointed out that the size looks smaller than the rest of the body text.

Quick fix is to increase the font-size a bit (to 15.5px) so we can improve readability and visual impact.

image

However, I'm partial to change the font altogether. I personally like ``Roboto condensed Light`` [Link]  (size: normal (15px))(https://fonts.google.com/specimen/Roboto+Condensed?preview.text=package,%20ansys-mapdl-core,%20pro&preview.text_type=custom):

image

Or even better ``consolas`` (size: normal (15px)):

image

### DIFF
--- a/pyansys_sphinx_theme/static/css/theme.css
+++ b/pyansys_sphinx_theme/static/css/theme.css
@@ -146,7 +146,7 @@
     color: #C89211;
     font-family: 'Inconsolata';
     font-weight: bold;
-    font-size: 15px;
+    font-size: 15.5px;
 }
 
 .sig-prename {
--- a/pyansys_sphinx_theme/static/css/theme.css
+++ b/pyansys_sphinx_theme/static/css/theme.css
@@ -145,6 +145,7 @@
     color: #C89211;
     font-family: "Source Code Pro", "Courier New", monospace;
     font-weight: bold;
+    font-size: 16px;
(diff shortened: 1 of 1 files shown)


germa89:
I acknowledge that half pixels will just be rounded, but it has some effects in some cases (zoom).

Link: https://stackoverflow.com/a/22832095/6650211

Andy-Grigg:
`consolas` is a great choice, and is my go-to font when I'm formatting monospace code in emails, Word, etc. But I think the issue will be cross-platform availability; `consolas` is a Microsoft-owned font and generally isn't available on non-Windows machines.

`Inconsolata` might be a good option, it's inspired by `consolas`, and is available for free on google fonts [here](https://fonts.google.com/specimen/Inconsolata#standard-styles), which is great, because that means we can include it in the CSS as a web font.

Font availability is always a bit of a pain. I really like [cascadia-code](https://github.com/microsoft/cascadia-code), but whilst it's an open source font, there's currently no official webfont support.

germa89:
Using ``Inconsolata`` together with ``@import`` and font-size 15.5px.

image

PR #20, opened 2022-01-28
Add breadcrumbs to all sphinx pages

### LINKED ISSUE #16: Add link to root of project documentation
From inside a project's documentation (e.g., https://dev.docs.pyansys.com/overview/index.html), there is no obvious way to get back to the project root (https://dev.docs.pyansys.com).

The logo in the top-left corner (which often takes this role) instead links to the root PyAnsys page https://docs.pyansys.com/.

I would argue that for a user looking at a *specific* page, it's more important to get to the root of that project, instead of the PyAnsys page. At the very least, there should be an obvious way of getting there.

Andy-Grigg:
Closes #16 

## Summary:

- Enable breadcrumbs with the `html_theme_option` `"show_breadcrumbs": True`
- The breadcrumbs appear on the page body itself as opposed to the navbar. I might have preferred to have them on the navbar so they are always visible, but I think that would require a major redesign of the theme
- The new css overrides some bootstrap defaults, so it was necessary to ensure it is added after the bootstrap css
- Additional 'root' breadcrumbs can be added, e.g. to the pyansys root page with the `html_theme_option` `"additional_breadcrumbs": [("Link text", "href")]`

## Issues:

- The breadcrumb that links to the module homepage has to be added manually when on a page within the docs. This is fine, but the problem is there's no way to get the title of the `index.rst` page. As a result, I'm relying on using the `html_title` as set in `conf.py`. The problem is that this isn't necessarily the same as the title of the `index.rst` page.

## Todo:

Assuming the approach is accepted, the docs need to be updated to describe how to enable this functionality and how to correctly set the `html_title` as described above.

## Screenshots

Docs without any changes to conf.py:

image

With breadcrumbs added: `    "show_breadcrumbs": True,`

image

With additional root element linking to pyansys: `"additional_breadcrumbs": [("PyAnsys Home", "https://docs.pyansys.com/")],`

image


akaszynski (approved):
This is excellent work and solves the navigation problem.

PR #22, opened 2022-02-01
Improve appearance and behavior of ReST `:download:` links

Andy-Grigg:
This is a relatively minor PR to improve how links generated with the ReST `:download:` directive work:

- CSS to theme them as hyperlinks instead of code blocks.
- JS to set them to open in a new tab by default. In my use case, this means download .py / .ipynb files instead of opening them in the current tab

Before / after screenshots:

image

image


pyansys_sphinx_theme/static/js/download_target_blank.js
```diff
@@ -0,0 +1,6 @@
+/* Add target="_blank" attribute to all hyperlinks generated by the ReST :download: directive.
+* This will ensure the links open in a new tab. */
+
+$(document).ready(function () {
```
germa89, line 4: Very clever!

germa89 (approved):
I think it is good. This is a very good idea and the result looks super nice. Honestly I'm not the most appropriate to judge your work, but I like it!

Andy-Grigg:
> I think it is good. This is a very good idea and the result looks super nice. Honestly I'm not the most appropriate to judge your work, but I like it!

Thanks! Do you want me to hold off until @akaszynski has a chance to review, or shall I merge?

germa89:
Up to you really, if you are in rush, just merge it. If not, surely @akaszynski  might have some useful contributions.

akaszynski:
Nothing to add. Thanks!

PR #31, opened 2022-03-28
Adding favicon to docs

germa89:
As mentioned in #983


akaszynski:
We need to have this made default for this theme. I'd include the icon instead in the theme assets directory and then make the icon default.

akaszynski (approved):

PR #51, opened 2022-05-25
Bump 0.3.dev0 version

### LINKED ISSUE #40: Development version should be 0.3.dev0
🐞 **Problem**

I just noticed that `_version.py` declares the current development version as "0.2.dev0". However, the "release/0.2" branch was already published. This implies that the developer's version should be "0.3.dev0" according to trunk-based development.

💡 **Solution**

Let us address this after #37, as the `_version.py` file is no longer going to be needed. It is better to have a single commit the usual "Bump 0.X.dev0 version" commit message. 

Once this gets done, we could release version "0.3.0", and quickly move to "0.4.dev0" so trunk-based development gets restored again.

Revathyvenugopal162:
fix #40

### DIFF
--- a/src/pyansys_sphinx_theme/__init__.py
+++ b/src/pyansys_sphinx_theme/__init__.py
@@ -1,7 +1,7 @@
 """This is the pyansys-sphinx-theme module."""
 import os
 
-__version__ = "0.2.dev0"
+__version__ = "0.3.dev0"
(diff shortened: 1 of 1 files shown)


jorgepiloto (approved):
Thanks for addresing this, @Revathyvenugopal162.

PR #61, opened 2022-06-01
Update labels

### LINKED ISSUE #60: Change PyPi label - pointing to wrong location
### Before submitting the issue

- [X] I have searched among the existing issues
- [X] I am using a Python virtual environment

### Description of the bug

Just change form ansys-templates to ansys-sphinx-theme

### Steps To Reproduce

Not needed

### Which Operating System are you using?

Linux

### Which Python version are you using?

3.9

### Installed packages

Not needed

RobPasMue:
Closes #60

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -10,7 +10,7 @@ Ansys Sphinx Theme
    :target: https://pypi.org/project/ansys-sphinx-theme/
    :alt: Python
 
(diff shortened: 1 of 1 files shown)


RobPasMue:
Current looks:

image

Shouldn't we just use a label like this one for the Python label? I find it way too verbose to specify all Python versions from 3.7 to 3.10:

image


If you agree I can change it in this PR

jorgepiloto (approved):
Good catch, @RobPasMue!

PR #75, opened 2022-06-14
add light theme config to css

### LINKED ISSUE #74: Dark and light mode coming from  pydata-sphinx-theme 0.9.0
We should try and block the dark/light theme duality since it has caused our docs to adopt different styles and colors don't match now. See below:
image
image

We should try and block this new feature so that it is only shown in our format... or adapt the logos so that the color is adapted at least... But I'd definitely block the dark/light mode (at least for now)

Revathyvenugopal162:
Fix #74 
Screenshot (19)
Screenshot (20)

### DIFF
--- a/src/ansys_sphinx_theme/static/css/ansys_sphinx_theme.css
+++ b/src/ansys_sphinx_theme/static/css/ansys_sphinx_theme.css
@@ -97,7 +97,7 @@ html[data-theme="dark"] {
   --pst-color-border: rgb(192, 192, 192);
   --pst-color-shadow: var(--pst-color-background);
   --pst-color-background: rgb(18, 18, 18);
-  --pst-color-on-background: rgb(30, 30, 30);
+  --pst-color-on-background: rgb(0, 0, 0);
   --pst-color-surface: rgb(41, 41, 41);
   --pst-color-on-surface: rgb(55, 55, 55);
 


RobPasMue:
Light theme works fine, but would there be a way to set the same "black" color for the logo and theme in dark mode?

akaszynski:
> Light theme works fine, but would there be a way to set the same "black" color for the logo and theme in dark mode?

I think this should be added before merging this PR.

germa89:
I wonder if it is better edit the CSS or the logo image.....

Revathyvenugopal162:
We can change the color. I will update it now. For dark theme giving another logo is also possible.

RobPasMue:
> 

Don't worry about editing the logo. In fact, the logo should stick to the official Ansys colours, Thus, I do agree with @germa89, its better editing the CSS files for the dark theme format. Thanks @Revathyvenugopal162

germa89 (approved):
Without checking the color values, it looks good to me. Looking forward to see the output online.

RobPasMue:
image


Looks awesome @Revathyvenugopal162 ! Thanks for solving the problem

RobPasMue (approved):

germa89:
I like the output a lot! The dark theme is much colorful though. But I like it that way. Good job @Revathyvenugopal162 !

germa89:
I like the output a lot! The dark theme is much colorful though. But I like it that way. Good job @Revathyvenugopal162 !

PR #85, opened 2022-07-25
Change the color of output cell for sphinx-gallery

Revathyvenugopal162:
This PR will fix the code output cell color of sphinx gallery. in dark mode. Light mode color retained as it is. 
out-dark
## Note :

The shade/color of texts and the background will change soon as discussed with documentation team once they are ready with the color codes.

### DIFF
--- a/src/ansys_sphinx_theme/static/css/ansys_sphinx_theme.css
+++ b/src/ansys_sphinx_theme/static/css/ansys_sphinx_theme.css
@@ -2,6 +2,7 @@
 @import "../basic.css";
 @import url('https://fonts.googleapis.com/css2?family=Inconsolata:wght@600&display=swap');
 @import "../tabs.css";
+@import "../sg_gallery.css";
 
 
 @font-face {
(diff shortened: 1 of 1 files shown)


src/ansys_sphinx_theme/static/css/ansys_sphinx_theme.css
```diff
@@ -2,6 +2,7 @@
 @import "../basic.css";
 @import url('https://fonts.googleapis.com/css2?family=Inconsolata:wght@600&display=swap');
 @import "../tabs.css";
+@import "../sg_gallery.css";
```
germa89, line 5:
```suggestion
@import "../sg_gallery.css";
```

From this line, I'm not sure if a file called `sg_gallery.css` should be in the ``_static`` folder or not. I am no expert in CSS or HTML, so it is your call.

germa89 (approved):
Just one question. Other than that, it looks good to me! Thank you for implementing this so fast!

src/ansys_sphinx_theme/static/css/ansys_sphinx_theme.css
Revathyvenugopal162: Yes if we are using sphinx_gallery and importing the package, then it will be there in _static folder.

MaxJPRey:
LGTM.

PR #92, opened 2022-08-10
Fix bold to ``<strong>`` tags

### LINKED ISSUE #79: Bold not showing bold
I am building my docs locally, using main branch (not release).

On the following code in a DOCSTRING (it seems to work in rst files, examples for instance):

```rst
            * ``STATIC`` or ``0`` - Perform a **static analysis**.  
              Valid for all degrees of freedom.

            * **BUCKLE or 1** - Perform a buckling analysis.  
              Implies that a previous static solution was performed with
              prestress effects calculated (``PSTRES,ON``).  
              Valid for structural degrees of freedom only.
```

I get:

image

Inspecting the web, it does make the text strong:
image

So I believe that maybe this bug was introduced in https://github.com/ansys/ansys-sphinx-theme/pull/75 
I have no probes though. 

I would not expect rendering differences between rst docstrings and rst files.

Revathyvenugopal162:
``<strong>``  tags are set to bolder but it was not showing bold. Font-weight: bolder means "One relative font weight heavier than the parent element" ([MDN article](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#values)). So giving ``strong`` tag to font_weight : 900 directly. Fix #79 
bold
bold-render

### DIFF
--- a/src/ansys_sphinx_theme/static/css/ansys_sphinx_theme.css
+++ b/src/ansys_sphinx_theme/static/css/ansys_sphinx_theme.css
@@ -554,6 +554,12 @@ html[data-theme="light"] .highlight .o {
   font-weight: bold;
 }
 
+/*
+#############################
+Bold font weight for **code**
+#############################
+*/
+
 b, strong {
   font-weight: 900;
 }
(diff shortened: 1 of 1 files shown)


germa89 (approved):
Looks good to me. Thank you @Revathyvenugopal162 !! :)

MaxJPRey (approved):

MaxJPRey:
LGTM!

PR #99, opened 2022-08-16
Fix left side panel active and hover color

### LINKED ISSUE #98: Hover color makes text dissapear
I really like the recent changes with the CSS, but the hover color (for the side panel on the left) is white for the light theme, which makes text disappear when the cursor hovers over it.

For reference, here's the element:
```
<a class="reference internal" href="packaging.html">
   Packaging
  </a>
```

Revathyvenugopal162:
Fix #98 by changing the hovering color to blue(link color) and active color to black ( with bold).
ansys-sphinx-theme left hover


RobPasMue (approved):
LGTM! This will require a patch release

MaxJPRey (approved):

MaxJPRey:
Thanks @Revathyvenugopal162 for the quick fix. As @RobPasMue , please do a patch release for this improvement.

PR #127, opened 2022-09-28
Add style for sphinx design

Revathyvenugopal162:
Add sphinx design styles to css files.


MaxJPRey (approved):
Perfect. Please check the rendering of the tab if you do any additional modifications but it should be OK.

jorgepiloto (approved):
This is great, @Revathyvenugopal162. Thanks for updating this 🚀

PR #130, opened 2022-10-05
Quick fix for the left panel size

Revathyvenugopal162:
The left panel is not rendering correctly, opened issue in dev guide(https://github.com/pyansys/dev-guide/issues/191), try to fix by rendering the left panel content by the next line, if line is long.

### DIFF
--- a/src/ansys_sphinx_theme/static/css/ansys_sphinx_theme.css
+++ b/src/ansys_sphinx_theme/static/css/ansys_sphinx_theme.css
@@ -531,12 +531,8 @@ Side column size (first and second column from left)
 */
 
 .col-md-3 {
-  flex: none;
-  max-width: fit-content;
-}
-
-.col-xl-7 {
-  max-width: fit-content;
+    flex: 0 0 20%;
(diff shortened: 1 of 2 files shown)


src/ansys_sphinx_theme/static/css/ansys_sphinx_theme.css
```diff
@@ -531,12 +531,8 @@ Side column size (first and second column from left)
...
 
 .col-md-3 {
-  flex: none;
-  max-width: fit-content;
-}
-
-.col-xl-7 {
-  max-width: fit-content;
+    flex: 0 0 20%;
+    max-width: 20%;
```
MaxJPRey, line 535: Let's keep an eye on it because in the past having the "size" of a panel based on percentage was the cause of some issues.

RobPasMue: Yeah... PyGeometry suffered of this issue with the percentage... Let's see

RobPasMue (approved):

PR #134, opened 2022-10-10
Modify font and scroll bar

Revathyvenugopal162:
- [ ] Add scroll bar visible
- [ ] Modify Font format 
scroll bar


MaxJPRey (approved):
@Revathyvenugopal162 This should make the navigation across our APIs easier. Thanks.

PR #150, opened 2022-11-16
Fix version switcher styling

Revathyvenugopal162:
Fix broken version switch styling.

### DIFF
--- a/src/ansys_sphinx_theme/static/css/ansys_sphinx_theme.css
+++ b/src/ansys_sphinx_theme/static/css/ansys_sphinx_theme.css
@@ -805,7 +805,7 @@ html[data-theme="dark"] .highlight pre {
 }
 
 #version_switcher_menu a.list-group-item.active {
-  background-color: var(--pst-color-background);
-  border-color: var(--pst-color-border);
-  color: var(--pst-color-text-base)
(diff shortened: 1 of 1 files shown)


MaxJPRey (approved):
LGTM.

src/ansys_sphinx_theme/static/css/ansys_sphinx_theme.css
```diff
@@ -805,7 +805,12 @@ html[data-theme="dark"] .highlight pre {
 }
 
 #version_switcher_menu a.list-group-item.active {
-  background-color: var(--pst-color-background);
-  border-color: var(--pst-color-border);
-  color: var(--pst-color-text-base)
+  background-color: #D09735!important;
```
jorgepiloto, line 808: Do we really want to use this color, see https://www.color-hex.com/color/d09735.

I am not sure if this is in conflict with our style guidelines. I think it works fine for the dark theme, but for the light one it may not be the best.

Revathyvenugopal162: No, it is not conflicting IMO, since the navigation bar is always black (in dark and in light) and it is only at the time of hovering, if you are putting the dropdown. But i am open to change.

PR #153, opened 2022-11-23
Migrate to pyansys/actions and include get_version_match function

Revathyvenugopal162:
Fix  as per discussion https://github.com/pyansys/dev-guide/pull/211#discussion_r1026182245, default position for the version switcher as per the image.

version-switcher

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -22,6 +22,15 @@ copyright = f"(c) {datetime.now().year} ANSYS, Inc. All rights reserved"
 author = "Ansys Inc."
 release = version = __version__
 
+
+def get_version_match(semver):
+    """Evaluate the version match for the multi-documentation."""
+    if semver.endswith("dev0"):
+        return "dev"
+    major, minor, _ = semver.split(".")
+    return ".".join([major, minor])
+
+
 # use the default ansys logo
 html_logo = ansys_logo_black
 html_theme = "ansys_sphinx_theme"
@@ -38,6 +47,10 @@ html_theme_options = {
     "github_url": "https://github.com/ansys/ansys-sphinx-theme",
     "use_edit_page_button": True,
     "contact_mail": "pyansys.support@ansys.com",
+    "switcher": {
+        "json_url": "https://raw.githubusercontent.com/ansys/ansys-templates/gh-pages/release/versions.json",  # noqa: E501
+        "version_match": get_version_match(__version__),
+    },
     "additional_breadcrumbs": [
         ("Ansys Internal Developer Portal", "https://dev.docs.ansys.com"),
     ],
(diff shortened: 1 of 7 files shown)


MaxJPRey (approved):
LGTM.

doc/source/conf.py
```diff
@@ -38,6 +47,10 @@
     "github_url": "https://github.com/ansys/ansys-sphinx-theme",
     "use_edit_page_button": True,
     "contact_mail": "pyansys.support@ansys.com",
+    "switcher": {
+        "json_url": "https://raw.githubusercontent.com/ansys/ansys-sphinx-theme/gh-pages/release/versions.json",  # noqa: E501
```
jorgepiloto, line 51: Let us use here a variable for the CNAME:

```suggestion
        "json_url": f"https://{cname}/release/versions.json",
```

jorgepiloto (changes requested):
Thanks for implementing this, @Revathyvenugopal162. I see that you took advantage of the pyansys/actions, nice!

Just left some suggestions about the way we manipulate the CNAME to make sure that it is only defined in one place.

Apart from those, everything looks great!

doc/source/conf.py
```diff
@@ -22,6 +22,15 @@
 author = "Ansys Inc."
 release = version = __version__
 
+
```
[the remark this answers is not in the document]
RobPasMue: I'm going to cry of happiness... My ideas are starting to be used elsewhere  😄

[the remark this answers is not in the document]
MaxJPRey: This is an excellent idea. I will try to use it too.

src/ansys_sphinx_theme/__init__.py
```diff
@@ -31,6 +30,14 @@ def get_html_theme_path():
     return Path(__file__).parents[0].absolute()
 
 
+def get_version_match(semver):
```
jorgepiloto, line 33: I really like you included this function here. This should be available by default to all projects using the template and allows us to forget about adding it manually. Thanks @Revathyvenugopal162.

src/ansys_sphinx_theme/__init__.py
```diff
@@ -38,6 +38,22 @@ def get_version_match(semver):
     return ".".join([major, minor])
 
 
+def prepare_html_configration(app, pagename, templatename, theme_options, doctree):
```
jorgepiloto, line 41: Do we really need to force the visibility of the switcher?

As of now, unless you enable it through your configuration, it will not be displayed. Also, forcing its appearance requires to impose a name to be displayed on it. Can we get the version number from the `app` parameter?

jorgepiloto (changes requested):

RobPasMue:
@Revathyvenugopal162 remember that if you change "Required" stages of your workflow, you have to change them in the settings of the repository. Otherwise the checks won't be green. This basically applies to the ``Doc building`` and ``Doc style`` stages, which replace the ``Documentation`` and ``Documentation style checks`` that were previously implemented.

PR #155, opened 2022-11-24
MAINT: inherit from pydata-sphinx-theme==0.12.0

### LINKED ISSUE #183: Update to `pydata-sphinx-theme==0.12.0`.
Update to pydata-sphinx-theme==0.12.0

Revathyvenugopal162:
Fix #183


jorgepiloto:
There seems to be an issue with the JSON builder. I'm running some debugging commits just in case.

jorgepiloto:
I just restored the "pydata-sphinx-version" to check if this is caused by the new versions. I am not getting the error locally.

jorgepiloto:
This is caused by the version. We need to identify for which version does the JSON builder break. If we do not find a workaround for this, either we add an option in pyansys/actions to specify desired builders or we restore our local doc-build job to only render HTML and PDF.

jorgepiloto:
However, I am afraid that this spreads to our projects and they can no longer build JSON.

Revathyvenugopal162:
> However, I am afraid that this spreads to our projects and they can no longer build JSON.

The problem is there building the pdf, not html
We need to rewrite that. are you building pdf with this workflow in any other lib @jorgepiloto ?

jorgepiloto:
Documentation fails for py310 + pydata-sphinx-theme==0.10.1. Let's try with py39.

jorgepiloto:
I bet the issue is with the JSON. From the CI log, you can see that the PDF has errors but those are ignored (due to the logic we apply in our Makefile) and that the JSON builder starts but then fails.

jorgepiloto:
Alright, I am able to reproduce this issue locally. It is the JSON builder.

jorgepiloto:
This issue needs to be fixed in Sphinx, I am opening a pull-request there and linking to this one.

Revathyvenugopal162:
> This issue needs to be fixed in Sphinx, I am opening a pull-request there and linking to this one

The issue is with the function in functools used by pydata-sphinx-theme.

https://github.com/sphinx-doc/sphinx/issues/10990#issuecomment-1327275982 we are able to find the issue, and let's wait until pydata fix this.

RobPasMue:
So... are we just tackling in this PR the update to pydata-sphinx-theme 0.12.0 or are we also fixing our CSS-related issues?

RobPasMue (approved):
In that case, LGTM!

MaxJPRey (approved):
LGTM.

germa89:
Thank you guys! I can close now https://github.com/pyansys/pymapdl/pull/1756

PR #161, opened 2022-11-25
Feat: Add default announcement

Revathyvenugopal162:
Add announcement layout, only if announcement file present.
Add layout and styles.


src/ansys_sphinx_theme/announcements.html
```diff
@@ -0,0 +1 @@
+<p>You are not reading the most recent version of this documentation. Latest stable release is <a href="https://actions.pyansys.com/release/1.0">1.0</a></p>
```
jorgepiloto, line 1: This file is not required. It can only live in the `gh-pages` of the repository.

src/ansys_sphinx_theme/announcement.html
```diff
@@ -0,0 +1,16 @@
+{% if theme_switcher %} 
+<div class="container-xl">
+    <div class="row"></div>
+        <div class="col-12 col-md-10" id="header-announcement">
+        <div class="col-12 col-md-13 col-xl-7" id="header-announcement-spacer">
+            <p id="announce"></p>
+            <script>
+            let file = "announcement.html"
```
jorgepiloto, line 8: I was just thinking about the file type for the announcement. Do you think we should read from an `announcement.txt` file? That way we allow the theme to take the control of the whole CSS.

jorgepiloto: Also, notice that this file is named `announcement.html` and this is reading from `announcement.html`, which causes an error. Could we rename this file like `announcement_layout.html`? This should be applied in the `layout.html` too.

src/ansys_sphinx_theme/layout.html
```diff
@@ -24,6 +24,12 @@
     {% endif %}
 {% endblock %}
 
+
+{% block extrahead %}
+    {% include 'announcement.html' %}
```
jorgepiloto, line 29: As per previous comments:

```suggestion
    {% include 'announcement_layout.html' %}
```

jorgepiloto (changes requested):

src/ansys_sphinx_theme/announcement.html
Revathyvenugopal162: Both should work, txt or html format will work, anyway we are just replacing the content. so doesn't matter, if it is html, dont add the <p> first but we can add the href, bold etc in html file, which is an advantage.

src/ansys_sphinx_theme/announcements.html
Revathyvenugopal162: Sure, this i added just for testing and we should remove it before merging.

src/ansys_sphinx_theme/announcement.html
Revathyvenugopal162: But thought of going with txt, because the developers can make custom announcement by just editing the txt file easily.

src/ansys_sphinx_theme/announcement_layout.html
```diff
@@ -0,0 +1,16 @@
+{% if theme_switcher %} 
+<div class="container-xl">
+    <div class="row"></div>
+        <div class="col-12 col-md-10" id="header-announcement">
+        <div class="col-12 col-md-13 col-xl-7" id="header-announcement-spacer">
+            <p id="announce"></p>
+            <script>
+            let file = "announcement.txt"
```
jorgepiloto, line 8: I just realized that it is not possible to create a link through a "txt" file in an easy way. Thus, let us revert this to an HTML file. Sorry fort the mess:

```suggestion
            let file = "announcement.html"
```

jorgepiloto (changes requested):

src/ansys_sphinx_theme/announcement_layout.html
jorgepiloto: I am not optimistic about this file path neither... What happens when this gets included in the header of a file that lives in a higher depth than the "announcement.html" file? I bet it does not find it and thus, does not include its content.

src/ansys_sphinx_theme/announcement_layout.html
```diff
@@ -0,0 +1,16 @@
...
+    <div class="row"></div>
+        <div id="header-announcement-spacer"></div>
+        <div class="col-lg-12 col-sm-6" id="header-announcement">
+            <p id="announce"></p>
+            <script>
+            let file = "announcement.html"
+            fetch (file)
+            .then(res => res.text())
+            .then(y => document.getElementById("announce").innerHTML = y);
+            </script>
```
jorgepiloto, line 12: This script should:

* Check if the file exists.
* If not, do not add any content.
* Ideally, use relative path for the file.

src/ansys_sphinx_theme/announcement_layout.html
```diff
@@ -0,0 +1,14 @@
+{% if theme_switcher %}
```
jorgepiloto, line 1: This logic is working, @Revathyvenugopal162. If no file is found, it will just print a message in the console but will not break the style.

I will update the version mapper to recursively add the `announcement.html` inside each folder of a release that is not the static one. This should do the job.

Could you please update the CSS file to apply for the `id="announce"`?

See how it is rendered now: https://templates.ansys.com/release/0.1/index.html

jorgepiloto (changes requested):

Revathyvenugopal162:
https://templates.ansys.com/release/0.1/getting_started/index.html please see the page for check the styles.

jorgepiloto (changes requested):
As per private discussion, let us change the announcement color from blue to lightred. Once done, this can be merged.

MaxJPRey (approved):
This is great.

jorgepiloto:
After UX team approval, this can be merged, @Revathyvenugopal162. Feel free to do so.

PR #182, opened 2023-01-23
Add the deprecation styles

### LINKED ISSUE #121: Problem with deprecated and versionadded notes color mode
Hi there! 
I was testing the different sphinx notes available to put in dosctrings, and I wanted to check `.. deprecated::` and `.. versionadded::` (see [here](https://github.com/pyansys/pydpf-core/blob/0249a141630b772b1fb1dd469015b7437bb7a53f/ansys/dpf/core/string_field.py#L24-L57)).

I found out that the general color theme is problematic as my local computer is in dark mode, giving this in doc dark mode:
image

Yet not rendering those two notes properly when in light or auto mode:
image
image

I think those two notes always use my local computer's dark mode despite the doc's mode.
The text is there, it is just not visible because the documentation's color theme makes it black, while the back of the note is not made white and seems to bypass the doc's color mode.
The problem might come from ansys-sphinx-theme, or it might come from our conf.py at PyDPF-Core.
What do you think?

Revathyvenugopal162:
Fix #121 

## Dark theme
depre-dark

## Light theme
ligh depre

### DIFF
--- a/doc/source/getting_started/index.rst
+++ b/doc/source/getting_started/index.rst
@@ -28,7 +28,7 @@ Consider using the ``conf.py`` for this repository:
 
 .. literalinclude:: ../conf.py
    :language: python
-   
+
 .. toctree::
    :hidden:
    :maxdepth: 2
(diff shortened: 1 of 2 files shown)


RobPasMue (approved):

MaxJPRey (approved):
LGTM.

PR #196, opened 2023-02-07
Modify workflow for dev docs deployment.

Revathyvenugopal162:
Modify dev docs deploy action to deploy on push to main.


MaxJPRey (approved):

RobPasMue (approved):

RobPasMue:
We should try to implement it like this everywhere... that way every time you run the workflow on main, you get an updated version of the docs. Thanks @Revathyvenugopal162 - this is def a good practice.

PR #208, opened 2023-02-22
Fix: The priority of workflow.

Revathyvenugopal162:
Fix the actions.

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -90,14 +90,6 @@ jobs:
     needs: build-library
     runs-on: ubuntu-latest
     steps:
-
-      - name: "Release to the private PyPI repository"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_cd.yml
```diff
@@ -53,7 +53,7 @@ jobs:
     name: Doc dev version deploy
     if: ${{ github.ref == 'refs/heads/main' }}
```
jorgepiloto, line 54: This will avoid overriding the development documentation with the latest stable release:

```suggestion
    if: github.event_name == 'push' && !contains(github.ref, 'refs/tags')
```

jorgepiloto (changes requested):

jorgepiloto (approved):

PR #211, opened 2023-02-22
Fix: misplaced styles of the meilisearch.

Revathyvenugopal162:
- Fix the search bar height.
- Fix the distance between search bar and dropdown button


jorgepiloto (approved):
Working fine both in Firefox and Chromium with various zoom levels.

PR #214, opened 2023-02-23
FIX : Make the switcher optional

Revathyvenugopal162:
Change the default switcher. Make the navigation end bar switcher only if it is present.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -11,13 +11,14 @@ from sphinx.builders.latex import LaTeXBuilder
 
 LaTeXBuilder.supported_image_types = ["image/png", "image/pdf", "image/svg+xml"]
 
-from ansys_sphinx_theme import (  # get_version_match,
+from ansys_sphinx_theme import (
     __version__,
     ansys_favicon,
     ansys_logo_black,
     ansys_logo_white,
     ansys_logo_white_cropped,
     generate_404,
+    get_version_match,
     latex,
     watermark,
 )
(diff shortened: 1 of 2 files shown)


doc/source/conf.py
```diff
@@ -60,10 +59,10 @@
...
         },
     ],
-    "switcher": {
-        "json_url": f"https://{cname}/release/versions.json",
-        "version_match": get_version_match(__version__),
-    },
+    # "switcher": {
+    #    "json_url": f"https://{cname}/release/versions.json",
+    #    "version_match": get_version_match(__version__),
+    # },
```
RobPasMue, line 65: Do not leave commented out code. If it is no longer needed, delete it.
```suggestion
```

doc/source/conf.py
```diff
@@ -11,14 +11,13 @@
 
 LaTeXBuilder.supported_image_types = ["image/png", "image/pdf", "image/svg+xml"]
 
-from ansys_sphinx_theme import (
+from ansys_sphinx_theme import (  # get_version_match,
```
RobPasMue, line 14:
```suggestion
from ansys_sphinx_theme import (
```
Same here

RobPasMue (changes requested):

doc/source/conf.py
Revathyvenugopal162: Yes, i was testing it, to check it is failing or not.

RobPasMue (approved):
Sounds reasonable. Thanks @Revathyvenugopal162

PR #218, opened 2023-02-23
FIX: The search-bar-width

Revathyvenugopal162:


akaszynski:
@Revathyvenugopal162, the only way I can get this to consistently work is to just fix the size of the search bar elements.

This is likely to be an adhoc fix that persists, but IMO it's sufficient for our work as we now no longer need to worry about the position of the dropdown.

Some of the content of the dropdown might be lost, but I think it's long enough to show most of the current content and the user can always click it to reveal the entire content.

akaszynski (approved):
Good to go if you're fine with my changes. Recommending a v0.9.4 release.

PR #235, opened 2023-04-14
fix: sphinx version and update footer

Revathyvenugopal162:
- [x] Add the compatibility with sphinx>6
- [x] Add the footer as ansys-sphinx-theme version.
##
image
## Expected output 
image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,7 @@ dependencies = [
 [project.optional-dependencies]
 doc = [
     "numpydoc==1.5.0",
-    "Sphinx==5.3.0",
+    "Sphinx==6.1.3",
     "sphinx-copybutton==0.5.1",
     "sphinx-notfound-page==0.8.3",
     "sphinx-design==0.4.1",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ doc = [
     "Sphinx==5.3.0",
     "sphinx-copybutton==0.5.1",
     "sphinx-notfound-page==0.8.3",
-    "sphinx-design==0.3.0",
+    "sphinx-design==0.4.1",
     "bs4==0.0.1",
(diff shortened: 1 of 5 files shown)


akaszynski (approved):
Saw that `sphinx-design` finally supported Sphinx v6.0

Thanks!

MaxJPRey (approved):
Nice to see the the footer.

src/ansys_sphinx_theme/__init__.py
```diff
@@ -85,6 +85,29 @@ def setup_default_html_theme_options(app):
     app.config.html_theme_options.setdefault("collapse_navigation", True)
 
 
+def update_footer_theme(app: Sphinx, pagename: str, templatename: str, context, doctree) -> None:
```
MaxJPRey, line 88: If possible, can you add the type of the `doctree` parameter.

RobPasMue: And ``context``

RobPasMue (approved):
LGTM! Thanks @Revathyvenugopal162

pyproject.toml
```diff
@@ -25,15 +25,15 @@ classifiers = [
     "Programming Language :: Python :: 3",
 ]
 dependencies = [
-    "Sphinx>=4.2.0,<6",
+    "Sphinx>=4.2.0",
```
jorgepiloto, line 28: Did you check this is not causing any issue in the top navigation bar, @Revathyvenugopal162? I remember we imposed this condition due to this issue.

jorgepiloto (approved):

pyproject.toml
Revathyvenugopal162: Yes, it is fixed, i checked it. https://github.com/pydata/pydata-sphinx-theme/issues/1130 issue in pydata related to that.

PR #246, opened 2023-05-16
MAINT: Change the pyansys-actions to ansys-actions

Revathyvenugopal162:
rename the pyansys/actions to ansys/actions.


MaxJPRey (approved):
Thanks for the fix @Revathyvenugopal162

Issue #256, opened 2023-07-15
Make card dimensions adaptable to all devices  


klmcadams:
### Description of the modifications

The cards on the main pyansys page (https://docs.pyansys.com/version/stable/) are squished when viewed on a mobile phone (see attached). We could make it so it only shows two cards per row, and/or adjust the css to be flexible and adapt to the size of the device being used. 

![image](https://github.com/ansys/ansys-sphinx-theme/assets/58492561/062f5d5a-4218-4a88-87ef-71e47dba052d)

### Useful links and references

https://docs.pyansys.com/version/stable/

Revathyvenugopal162:
Thanks for opening issue @klmcadams , @AlejandroFernandezLuces already mentioned this issue, and i am working on this.

MaxJPRey:
@Revathyvenugopal162 When it is done, might be worth it to do some additional testing about other contents and pages. For instance, make sure that the dev guide is rendering properly and we did not downgraded any UI/UX part.

RobPasMue:
Testing on different devices might be difficult in terms of CI/CD. But we should explore this. I'm no expert on CSS but happy to learn

Revathyvenugopal162:
It is already doing with pydata, we do have an open issue to adapt tests, we are adding this with different devices when we are modifying with pydata, but for sphinx cards, they dont have customization, so we just have to adapt their tests only for the sphinx design, until they have some from their side.

Revathyvenugopal162:
Thank you for opening the issue, fixed with v1.0.5 and closing the issue.

PR #259, opened 2023-07-25
MAINT: Add sphinx card configuration according to screen size

Revathyvenugopal162:
Resolve part of #256 
image

- The sphinx card with content in footer is having the issue with  screen size, so adapted only for the card header 
and size. https://cheatsheets.docs.pyansys.com/ rendering very good on all screen.
- Update the script to render all examples of the card inside the example section for manual check and also as examples
- Adapt the icon with same font size
### 
image

image

### DIFF
--- a/doc/source/examples/sphinx-design.rst
+++ b/doc/source/examples/sphinx-design.rst
@@ -17,7 +17,7 @@ please refer `sphinx design <https://sphinx-design.readthedocs.io/en/latest/inde
         .. literalinclude:: sphinx_examples/{{ filename }}
            :language: rst
         
-    This directive renders the as follow:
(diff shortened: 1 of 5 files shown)


doc/source/examples/sphinx-design.rst
```diff
@@ -17,7 +17,7 @@ please refer `sphinx design <https://sphinx-design.readthedocs.io/en/latest/inde
         .. literalinclude:: sphinx_examples/{{ filename }}
            :language: rst
         
-    This directive renders the as follow:
+        This directive renders as follows:
```
Revathyvenugopal162, line 20:
```suggestion
This directive renders as follows:
```

PR #275, opened 2023-08-16
fix: rename the extension name with no conflict

### LINKED ISSUE #274: Handler error in latest version
### 🔍 Before submitting the issue

- [X] I have searched among the existing issues
- [X] I am using a Python virtual environment

### 🐞 Description of the bug

When running the docs I'm getting this error:

```output
writing output... [ 87%] mapdl_commands/solution/_autosummary/ansys.mapdl.core.Mapdl.bfv .. mapdl_commands/solution/_autosummary/ansys.mapdl.core.Mapdl.mfelem
Sphinx parallel build error:
sphinx.errors.ExtensionError: Handler <function update_and_remove_templates at 0x7f55445a41f0> for event 'html-page-context' threw an exception (exception: 'ansys' is not in list)
make: *** [Makefile:21: html] Error 2
```
https://github.com/ansys/pymapdl/actions/runs/5870157592/job/15917918928?pr=2248#step:15:231

### 📝 Steps to reproduce

It happens in PyMAPDL CICD.

### 💻 Which operating system are you using?

Linux

### 📀 Which ANSYS version are you using?

N/A

### 🐍 Which Python version are you using?

3.9

### 📦 Installed packages

```shell
NA
```

Revathyvenugopal162:
fix #274

### DIFF
--- a/src/ansys_sphinx_theme/__init__.py
+++ b/src/ansys_sphinx_theme/__init__.py
@@ -137,7 +137,7 @@ def fix_edit_html_page_context(
     see https://github.com/pyvista/pyvista/pull/4113
     """
 
-    def fix_edit_link_button(link: str) -> str:
+    def fix_edit_link_page(link: str) -> str:
         """Transform "edit on GitHub" links to the correct URL.
 
         This function fixes the URL for the "edit this page" link.
@@ -203,7 +203,7 @@ def fix_edit_html_page_context(
         else:
             return link
 
-    context["fix_edit_link_button"] = fix_edit_link_button
+    context["fix_edit_link_page"] = fix_edit_link_page
 
 
 def update_footer_theme(
--- a/src/ansys_sphinx_theme/theme/ansys_sphinx_theme/components/edit-this-page.html
+++ b/src/ansys_sphinx_theme/theme/ansys_sphinx_theme/components/edit-this-page.html
@@ -1,7 +1,7 @@
 {% if sourcename is defined and theme_use_edit_page_button==true and
 page_source_suffix %} {% set src = sourcename.split('.') %}
 <div class="tocsection editthispage">
-  <a href="{{ fix_edit_link_button(get_edit_provider_and_url()[1]) }}">
+  <a href="{{ fix_edit_link_page(get_edit_provider_and_url()[1]) }}">
     <i class="fa-solid fa-pencil"></i>
     {% set provider = get_edit_provider_and_url()[0] %} {% block
     edit_this_page_text %} {% if provider %} {% trans provider=provider %}Edit


germa89:
I don't see how the changes proposed can aliviate #274 ... Can you explain a bit please?

Revathyvenugopal162:
> I don't see how the changes proposed can alleviate #274 ... Can you explain a bit please?

i think the function 'fix_edit_link_button' is defined in ansys-sphinx theme and in pymapdl, made the jinja to add the function in html context, and shadowing  fix_edit_link_button function, leading to unexpected behavior of looking for an extension named ansys.

Revathyvenugopal162:
I tried it locally and it is fixing my error for pymapdl!

RobPasMue:
Oh... so it's because PyMAPDL has it's own function? In that case... we might want to change PyMAPDL rather than the Sphinx theme... right?

Revathyvenugopal162:
> Oh... so it's because PyMAPDL has it's own function? In that case... we might want to change PyMAPDL rather than the Sphinx theme... right?

Yes, but i think we both got that name from pyvista, if we are adding this as another extension or add theme in public themes, it should not conflict another library, so i changed it here.!
But changing for the moment in pymapdl will also solve the issue.

RobPasMue:
> Yes, but i think we both got that name from pyvista, if we are adding this as another extension or add theme in public themes, it should not conflict another library, so i changed it here.! But changing for the moment in pymapdl will also solve the issue.

Okay, let's accept the change then =)

RobPasMue (approved):

germa89 (approved):
With the recent explanations, I see the point for this PR and I approve it.

Looking forward to see this change in the next release. Thank a lot @Revathyvenugopal162 !! 😄

germa89:
Also, how did you discover what was the reason? It seems to me the error was a bit obscure. Very good job!

Revathyvenugopal162:
> By the way, I guess your fix for the "edit button" should also work with PyMAPDL I guess? So maybe we can remove that bit from PyMAPDL.

i believe, we can remove some code and the extension with these changes.

MaxJPRey (approved):
Thanks for the explaination @Revathyvenugopal162 .

germa89:
> > By the way, I guess your fix for the "edit button" should also work with PyMAPDL I guess? So maybe we can remove that bit from PyMAPDL.
> 
> i believe, we can remove some code and the extension with these changes.

I will.. eventually... one day.. find the time for this. 🥲

RobPasMue:
Can we merge this BTW?

germa89:
We will need a patch release for this.

RobPasMue:
Yep, @Revathyvenugopal162 - please plan on it

Revathyvenugopal162:
Sure, i will do this Morning.

PR #286, opened 2023-08-30
Fix:edit this context for undefined domain

Revathyvenugopal162:
Add try block, for the code extracts specific information (domain, signode data, etc.) and formats it into an info dictionary.

### DIFF
--- a/src/ansys_sphinx_theme/__init__.py
+++ b/src/ansys_sphinx_theme/__init__.py
@@ -183,7 +183,7 @@ def fix_edit_html_page_context(
                             github_version=kind,
                             edit=True,
                         )
-                except Exception as e:
+                except ValueError as e:
                     logging.debug(f"An error occurred: {e}")  # Log the exception as debug info
                     return link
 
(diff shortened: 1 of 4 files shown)


src/ansys_sphinx_theme/__init__.py
```diff
@@ -159,28 +160,32 @@ def fix_edit_link_page(link: str) -> str:
...
+                        # This is an API example
+                        return sphinx_linkcode_resolve(
+                            domain=domain,
+                            info=info,
+                            library=f"{github_user}/{github_repo}",
+                            source_path=github_source,
+                            github_version=kind,
+                            edit=True,
+                        )
+                except Exception as e:
```
RobPasMue, line 186: I mean... catching any exception is sometimes the "only" solution but... is there a way for you to narrow down what Exception is expected here?

RobPasMue (commented):
I'll approve it - though I'd prefer to narrow down the kinds of Exceptions we accept

RobPasMue (approved):

src/ansys_sphinx_theme/__init__.py
Revathyvenugopal162: yes it a value error raised by pydata

PR #297, opened 2023-09-18
fix: try except to handle value error for linkcode extension 

Revathyvenugopal162:
https://github.com/ansys/pymapdl/pull/2341 fixes the value error


germa89 (approved):
I dont really understand how this works.. but I trust you

germa89:
I checked and it seems to work locally in PyMAPDL.

PR #299, opened 2023-09-19
fix: update the extension if condition

### LINKED ISSUE #298: 'if' condition on 'uri' not needed
I think that this `if` condition is not needed as we are testing it few line below.

Or you could remove the second part of the condition statement below `or not uri`.

I might be missing a point though.
The purpose of this issue is to make sure that it is intended.

https://github.com/ansys/ansys-sphinx-theme/blob/d9fe77c8e71f480109ad0eed630e86289091d2a1/src/ansys_sphinx_theme/extension/linkcode.py#L237

Revathyvenugopal162:
fix #298

### DIFF
--- a/src/ansys_sphinx_theme/extension/linkcode.py
+++ b/src/ansys_sphinx_theme/extension/linkcode.py
@@ -227,6 +227,7 @@ def link_code(app: Sphinx, doctree: Node):
 
             # Call user code to resolve the link
             try:
+                # Resolve the URI for source code link
                 uri = sphinx_linkcode_resolve(
                     domain=domain,
                     info=info,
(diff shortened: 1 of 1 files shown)


src/ansys_sphinx_theme/extension/linkcode.py
```diff
@@ -227,19 +227,18 @@ def link_code(app: Sphinx, doctree: Node):
...
                     github_version=github_version,
                 )
-                if not uri:
-                    # no source
-                    continue
 
-                if uri in uris or not uri:
-                    # only one link per name, please
+                # Check if the URI is already processed
+                if uri in uris:
```
RobPasMue, line 240:
```suggestion
                if not uri or uri in uris:
```

Check that you have an object returned... and in case you have it, check that it is already in the uris list

If conditions are linear: you evaluate first the first statement, then the second statement... This will make it robust

germa89: My life changed when I got to know about that behaviour of the `and` operator.

germa89 (approved):

PR #300, opened 2023-09-19
fix: update the styles with pydata-sphinx-theme update

Revathyvenugopal162:


RobPasMue:
All these changes would only work with pydata-sphinx-theme>=0.14, right? We should update this in the pyproject.toml

RobPasMue (approved):

PR #308, opened 2023-09-29
fix : update the js compactible with require.js module loader

Revathyvenugopal162:
update the js code to work with the js module loader


jorgepiloto (approved):

RobPasMue (approved):

germa89:
Wondering if this will fix plotly support :)

PR #331, opened 2024-01-04
fix: search bar fail

Revathyvenugopal162:


germa89 (approved):
No very sure what is happening.. but I trust you :)

PR #334, opened 2024-01-11
Wip/test fix side bar

SMoraisAnsys:
⚠️ This is a PR whose content must be rewritten before it can be merged ⚠️ 

As it stands, the associated branch performs a change to be able to use the current main branch of https://github.com/pydata/pydata-sphinx-theme The idea here is to be able to use the fix that has been merged in 
 https://github.com/pydata/pydata-sphinx-theme/pull/1632 to fix the problem of side bare when using option :hidden:.

I've been able to validate this change through https://github.com/ansys-internal/pyansys-edb/pull/111

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,8 @@ classifiers = [
 ]
 dependencies = [
     "Sphinx>=4.2.0",
-    "pydata-sphinx-theme>=0.14.0",
+    "pydata-sphinx-theme@git+https://github.com/pydata/pydata-sphinx-theme#egg=main",
+    # "pydata-sphinx-theme>=0.14.0",
     "Jinja2>=3.1.2",
     "importlib-metadata>=4.0",
 ]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -26,7 +26,8 @@ classifiers = [
 ]
 dependencies = [
     "Sphinx>=4.2.0",
-    "pydata-sphinx-theme>=0.14.0",
+    "pydata-sphinx-theme@git+https://github.com/pydata/pydata-sphinx-theme#egg=main",
```
jorgepiloto, line 29: The issue with this is that using a branch does not allow to upload a new version to PyPI.

We need to limit the upper version for the `pydata-sphinx-theme`.

jorgepiloto (changes requested):
Pinging here @Revathyvenugopal162.

jorgepiloto:
I think the latest version to work for us was `pydata-sphinx-theme==0.14.4`

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/search.html
jorgepiloto, line 1: Let's discard these changes.

jorgepiloto (changes requested):

jorgepiloto:
Dependabot should be the one opening a new PR every time there is a new major release for the pydata-sphinx-theme. If so, we should review in depth the rendering of this new version.

SMoraisAnsys:
Closing this PR but keeping the branch alive ftm.

PR #335, opened 2024-01-12
fix: downgrade the pydata sphinx theme

Revathyvenugopal162:


pyproject.toml
```diff
@@ -26,7 +26,7 @@ classifiers = [
 ]
 dependencies = [
     "Sphinx>=4.2.0",
-    "pydata-sphinx-theme>=0.14.0",
+    "pydata-sphinx-theme==0.14.4",
```
clatapie, line 29: Should we be less restrictive by asking `<=0.14.4`?

RobPasMue: I'd agree here... pinning down a specific PyData Sphinx Theme version seems too much for me

RobPasMue, line 29:
```suggestion
    "pydata-sphinx-theme>=0.14,<0.15",
```

Something like this seems better suited

Revathyvenugopal162: if we are not giving the specified version, dependabot will not open a PR right? but its ok, we can keep an eye on their release

RobPasMue: Dependabot should open a PR on upper limits as well (if I remember properly)

RobPasMue (approved):

PR #355, opened 2024-04-03
fix: add additional dependencies for autoapi templates

Revathyvenugopal162:
Adds dependencies for API documentation generation with custom templates provided with ansys-sphinx-theme

### DIFF
--- a/doc/source/getting_started/index.rst
+++ b/doc/source/getting_started/index.rst
@@ -3,9 +3,9 @@
 ===============
 Getting started
 ===============
-`Sphinx <https://www.sphinx-doc.org/en/master/>`_ is a Python documentation
+`Sphinx <Sphinx_>`_ is a Python documentation
 generator for creating documentation. If you are new to using Sphinx, see
-`Sphinx Getting Started <https://www.sphinx-doc.org/en/master/usage/quickstart.html>`_.
+`Sphinx Getting Started <Sphinx_Getting_Started_>`_.
 
 This section explains how to install the Ansys Sphinx theme and then set up your
 Sphinx ``conf.py`` file to use this theme to generate your documentation.
@@ -16,9 +16,9 @@ Dependencies
 Ansys sphinx theme build on top of ``pydata sphinx theme``. 
 The theme requires the following dependencies:
 
-- `sphinx <https://www.sphinx-doc.org/en/master/>`_
-- `pydata-sphinx-theme <https://pydata-sphinx-theme.readthedocs.io/en/latest/>`_
-- `Jinja2 <https://jinja.palletsprojects.com/en/2.11.x/>`_
+- `Sphinx <Sphinx_PyPI_>`_
+- `pydata-sphinx-theme <PyData_PyPI_>`_
+- `Jinja2 <Jinja2_PyPI_>`_
 
 Optional dependencies
 ---------------------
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -28,10 +28,13 @@ dependencies = [
...
     "Jinja2>=3.1.2",
-    "importlib-metadata>=4.0",
 ]
 
 [project.optional-dependencies]
+autoapi = [
+    "sphinx-autoapi==3.0.0a4",
+    "sphinx-design==0.5.0",
+]
 doc = [
```
RobPasMue, line 38: I love this!

src/ansys_sphinx_theme/__init__.py
```diff
@@ -34,13 +34,7 @@
...
 from ansys_sphinx_theme.latex import generate_404  # noqa: F401
 
-try:
-    import importlib.metadata as importlib_metadata
-except ModuleNotFoundError:  # pragma: no cover
-    import importlib_metadata
-
-__version__ = importlib_metadata.version(__name__.replace(".", "-"))
-
+__version__ = "0.16.dev0"
```
RobPasMue, line 37: Why are we changing the version approach?

RobPasMue (approved):

src/ansys_sphinx_theme/__init__.py
Revathyvenugopal162: In updating the documentation for the dependencies, I've removed the importlib metadata dependencies when utilizing dynamic in pyproject.toml.  If there are any concerns or disagreements regarding this adjustment, I'm open to reverting these changes.

RobPasMue: Starting on Python 3.9, importlib.metadata comes by default with Python - which for our case means that you can drop the try/except and just perform the import. I would leave it as it was IMO, without the try/except. All repos are working the same way, plus I think it's cleaner to have the version in the project's configuration file rather than inside the source code itself.

RobPasMue: Also - in terms of scope - the title of the PR does NOT reflect this change so, in any case I'd do it in a different PR.

Revathyvenugopal162: ok noted, thank you and reverted the changes.

doc/source/getting_started/index.rst
```diff
@@ -10,6 +10,30 @@ generator for creating documentation. If you are new to using Sphinx, see
...
 
+Dependencies
+------------
+
+Ansys sphinx theme build on top of ``pydata sphinx theme``. 
+The theme requires the following dependencies:
+
+- `sphinx <https://www.sphinx-doc.org/en/master/>`_
+- `pydata-sphinx-theme <https://pydata-sphinx-theme.readthedocs.io/en/latest/>`_
+- `Jinja2 <https://jinja.palletsprojects.com/en/2.11.x/>`_
```
jorgepiloto, line 21: Since we are talking about dependencies here, I guess we could point to the official pages in PyPI. What do you think?

jorgepiloto (approved):

doc/source/getting_started/index.rst
Revathyvenugopal162: yeah sure, i will add those.

SMoraisAnsys (approved):

Issue #354, opened 2024-04-02
Revisit sphinx-autoapi templates


jorgepiloto:
Ensure that:

- [x] Object name is not included in method or property description. Can only be fixed via CSS to ensure xref are still valid and unique. Implemented in #358. Example:

      ```
      FooObject.bar_property -> bar_property
      ```
      
- [x] Create "api" group in external dependencies. Implemented in #355 
- [x] Use Nerd Fonts for rendering symbols in toctree. Implemented in #362
- [x] Ensure typehints are using "short" qualification when rendering. Example:

      ```
      def function(foo: ansys.library.core.Foo) -> def function(foo: Foo)
      ```
      
     Default "sphinx-autoapi" templates for "method.rst" are using the [Sphinx Python Domain](https://www.sphinx-doc.org/en/master/usage/domains/python.html) to avoid enabling additional extensions. Consider using "autodoc" rather than `.. py:method::`.
     
     See `typehints_fully_qualified = False` from https://github.com/tox-dev/sphinx-autodoc-typehints
     See `autodoc_typehints_format = "short"` from https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html #358

Revathyvenugopal162:
closing since All tasks have been completed.

PR #358, opened 2024-04-05
fix: Update ``autoapi`` style to for disabling ``prename`` display.

Revathyvenugopal162:
Corrected display issue where function and property prefixes inside a class were improperly shown, e.g., `FooObject` --> `bar_property` displayed as `FooObject.bar_property` in the `FooObject` class instead of `bar_property`. This PR disables prefix display to maintain consistency with the function name only.

### DIFF
--- a/src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/css/ansys_sphinx_theme.css
+++ b/src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/css/ansys_sphinx_theme.css
@@ -1161,3 +1161,9 @@ a:visited:hover {
 .bd-sidebar-primary {
   padding: 0.5rem;
 }
+
+/* Auto-api prename */
+
+.sig-prename.descclassname {
+  display: none;
+}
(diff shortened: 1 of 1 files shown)


src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/css/ansys_sphinx_theme.css
```diff
@@ -1161,3 +1161,9 @@ a:visited:hover {
 .bd-sidebar-primary {
   padding: 0.5rem;
 }
+
+/* Auto-api prename */
```
jorgepiloto, line 1165: Let's provide a more descriptive comment on what this is targeting at:

```suggestion
/* Do not display object domain namespace when documenting objects.

   Example: class ansys.module.foo.Foo.method -> method

  */
```

jorgepiloto (approved):
Awesome! This addresses part of issue #354. Please, always a description when opening a pull-request. Even if the title already describes it. It is always helpful to link issues.

RobPasMue (approved):
LGTM!

PR #362, opened 2024-04-08
feat: add nerd fonts for ``autoapi`` templates icon

Revathyvenugopal162:
add nerd fonts for autoapi templates icon


doc/source/conf.py
jorgepiloto, line 1: Why was dis file modified?

Revathyvenugopal162: I probably need to revert these changes. I made adjustments to this file because PyGitHub was repeatedly raising errors due to my numerous attempts to build the docs for testing purposes and pulling content. Therefore, I added my Personal Access Token (PAT) to address this issue

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/_templates/autoapi/python/module.rst
```diff
@@ -29,7 +29,7 @@
    :hidden:
 
     {% for obj in objects_list %}
-    {{ icon }} {{ obj.short_name }}<{{ obj.include_path }}>
+    <i class="{{ icon }}"></i> {{ obj.short_name }}<{{ obj.include_path }}>
```
jorgepiloto, line 32:
```suggestion
    <i class="{{ icon }}" /> {{ obj.short_name }}<{{ obj.include_path }}>
```

jorgepiloto: Also, is there any way the `<` and `>` do not render so we can avoid replacing the `&lt` and `&gt` with the actual symbols? What about escaping these?

Revathyvenugopal162: > Also, is there any way the `<` and `>` do not render so we can avoid replacing the `&lt` and `&gt` with the actual symbols? What about escaping these?

I'm testing it out and looking into other options too. I'll let you know if I find a solution.

src/ansys_sphinx_theme/__init__.py
```diff
@@ -110,9 +110,56 @@ def get_autoapi_templates_dir_relative_path(path: pathlib.Path) -> str:
...
+
+    def prepare_jinja_env(jinja_env) -> None:
+        """Prepare the Jinja environment for the theme."""
+        jinja_env.globals["project_name"] = autoapi_project_name
+
+    # Set the autoapi options
+    app.config["autoapi_prepare_jinja_env"] = prepare_jinja_env
+    app.config["autoapi_type"] = autoapi.get("type", "python")
+    app.config["autoapi_root"] = autoapi.get("output", "api")
+    app.config["autoapi_own_page_level"] = autoapi.get("own_page_level", "module")
```
jorgepiloto, line 156: This should be class by default:

```suggestion
    app.config["autoapi_own_page_level"] = autoapi.get("own_page_level", "class")
```

src/ansys_sphinx_theme/__init__.py
```diff
@@ -110,9 +110,56 @@ def get_autoapi_templates_dir_relative_path(path: pathlib.Path) -> str:
...
+    for extension in required_extensions:
+        if extension not in app.config["extensions"]:
+            app.config["extensions"].append(extension)
+    AUTOAPI_OPTIONS = [
+        "members",
+        "undoc-members",
+        "show-inheritance",
+        "show-module-summary",
+        "special-members",
+    ]
```
jorgepiloto, line 138: Is there a way of customizing these? Not all projects are using the `special-members`, for instance.

src/ansys_sphinx_theme/__init__.py
```diff
@@ -372,6 +419,26 @@ def download_file(url: str, directory: pathlib.Path) -> None:
...
+        Exception that occurred during the build process.
+    """
+    if exception is None:
+        build_dir = pathlib.Path(app.builder.outdir).resolve()
+        file_names = [str(file) for file in build_dir.rglob("*.html")]
+        for file_name in file_names:
+            with open(build_dir / file_name, "r", encoding="utf-8") as f:
+                content = f.read()
+            with open(build_dir / file_name, "w", encoding="utf-8") as f:
+                f.write(content.replace("&lt;", "<").replace("&gt;", ">"))
```
jorgepiloto, line 439: Some tips:

- Try to reduce indentation as much as possible. This can usually be achieved by checking the opposite condition.

  ```python
  if exception is not None:
      return
  ```

I would force this to run only in the HTML files located in an `api/` dir.

jorgepiloto (changes requested):

src/ansys_sphinx_theme/__init__.py
Revathyvenugopal162: > Some tips:
> 
> * Try to reduce indentation as much as possible. This can usually be achieved by checking the opposite condition.
>   ```python
>   if exception is not None:
>       return
>   ```
> 
> I would force this to run only in the HTML files located in an `api/` dir.

sure, thanks jorge, i will try to adapt this.

src/ansys_sphinx_theme/__init__.py
Revathyvenugopal162: yes , they can give these options as a list in theme options.

doc/source/user_guide/index.rst
```diff
@@ -13,7 +13,7 @@ While the Ansys Sphinx theme is often used as is, you can customize the followin
    :maxdepth: 2
    :includehidden:
 
-   css
+   \<span class="nf nf-md-package"\>\</span\> css <css.rst>
```
Revathyvenugopal162, line 16:
```suggestion
   css
```

jorgepiloto (approved):
Awesome!

PR #380, opened 2024-05-22
docs: update changelog_template.jinja

RobPasMue:
As title says


RobPasMue:
Might be worth a second patch release @Revathyvenugopal162

Revathyvenugopal162 (approved):

PR #400, opened 2024-06-14
fix: move the version switcher to left navigation bar

Revathyvenugopal162:
image

### DIFF
--- a/doc/changelog.d/400.fixed.md
+++ b/doc/changelog.d/400.fixed.md
@@ -1,0 +1,1 @@
+fix: move the version switcher to left navigation bar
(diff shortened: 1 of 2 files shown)


RobPasMue:
Looks a bit odd, doesn't it? @ansys/pyansys-core

RobPasMue:
Or maybe it's just me that I am already used to seeing it in the right top corner 😄

SMoraisAnsys:
> Or maybe it's just me that I am already used to seeing it in the right top corner 😄

I'm also more familiar with seeing this at the top right corner (not only from pyansys documentations but other projects, e.g. https://matplotlib.org/stable/)

germa89:
I agree with the other comments, right side looks better.

Revathyvenugopal162:
In that case, i will close the PR and remove this from the meta issue.

jorgepiloto:
Closing this as per the style guidelines suggested in #403

PR #401, opened 2024-06-17
fix: update the github icon

Revathyvenugopal162:
## before
image

## after
image

### DIFF
--- a/src/ansys_sphinx_theme/__init__.py
+++ b/src/ansys_sphinx_theme/__init__.py
@@ -176,8 +176,7 @@ def setup_default_html_theme_options(app):
     theme_options.setdefault("collapse_navigation", True)
     theme_options.setdefault("navigation_with_keys", True)
 
-    # update github icon links
-    # if icon links are not given override them with default links
+    # Update the icon links. If not given, add a default GitHub icon.
     if not theme_options.get("icon_links") and theme_options.get("github_url"):
         theme_options["icon_links"] = [
             {
--- a/src/ansys_sphinx_theme/__init__.py
+++ b/src/ansys_sphinx_theme/__init__.py
@@ -167,12 +167,26 @@ def setup_default_html_theme_options(app):
 
     """
     # Place all switchers and icons at the end of the navigation bar
-    if app.config.html_theme_options.get("switcher"):
-        app.config.html_theme_options.setdefault(
+    theme_options = app.config.html_theme_options
+    # Place all switchers and icons at the end of the navigation bar
+    if theme_options.get("switcher"):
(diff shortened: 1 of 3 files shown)


src/ansys_sphinx_theme/__init__.py
```diff
@@ -167,12 +167,26 @@ def setup_default_html_theme_options(app):
...
+        theme_options.setdefault(
             "navbar_end", ["version-switcher", "theme-switcher", "navbar-icon-links"]
         )
-    app.config.html_theme_options.setdefault("collapse_navigation", True)
-    app.config.html_theme_options.setdefault("navigation_with_keys", True)
+    theme_options.setdefault("collapse_navigation", True)
+    theme_options.setdefault("navigation_with_keys", True)
+
+    # update github icon links
+    # if icon links are not given override them with default links
```
jorgepiloto, line 180:
```suggestion
    # Update the icon links. If not given, add a default GitHub icon.
```

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/css/ansys_sphinx_theme.css
```diff
@@ -600,6 +600,10 @@ Navigation column (according to side column)
   display: unset;
 }
 
+.theme-switch-button span {
+  color: #fff;
```
germa89, line 604: Wouldn't it be better to use the ``--pst-color-primary`` variable? So you "forget about it" in the next theme changes?

germa89 (approved):
LGTM.

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/css/ansys_sphinx_theme.css
Revathyvenugopal162: I am using these variables for all items other than the navigation bar, which remains black (until we change its color). The primary color is different for both theme options, but we can define it to match the navigation color.

jorgepiloto: Even with that, we should have a variable for basic colors. These could be later reused by the `--pst-color-primary` when declared for the light and dark themes.

For the moment, we can leave it as this. We are going to reformat the CSS very soon.

jorgepiloto (approved):

MaxJPRey (approved):
LGTM.

Revathyvenugopal162:
Thank you all for the review.

PR #407, opened 2024-06-25
refactor: remove function duplicate

SMoraisAnsys:
As title says, the function is defined twice with the same content.


Revathyvenugopal162 (approved):
LGTM, Thanks @SMoraisAnsys

MaxJPRey (approved):
LGTM.

Issue #403, opened 2024-06-18
Ansys Sphinx Theme - Design updates


mia-guo-ux:
**Sphinx design** : https://sphinxdocs.ansys.com/version/dev/examples/sphinx-design.html#

The following components could be enhanced to create a more streamlined user experience. 
- Cards (including card carousel)
- Buttons 
- Links
- Side navigation
- Selector 
- Dropdown 
- Table
- etc.

Please feel free to let me know if you have any concerns or if anything needs to be specified. Cheers!
FYI @Revathyvenugopal162 @jorgepiloto @RobPasMue

mia-guo-ux:
**Card**
![Home](https://github.com/ansys/ansys-sphinx-theme/assets/123198029/20ceb8fd-0542-4b2f-a555-e6b0bd931e2c)

mia-guo-ux:
**Buttons**
![Buttons](https://github.com/ansys/ansys-sphinx-theme/assets/123198029/6e76a1a0-5ff4-4c0a-8486-2646864015fe)

mia-guo-ux:
**Links**
![Links](https://github.com/ansys/ansys-sphinx-theme/assets/123198029/7a750571-5980-43fa-81ac-e722e6d19a9e)

mia-guo-ux:
**Side nav**
![Side nav](https://github.com/ansys/ansys-sphinx-theme/assets/123198029/de6f591a-2c80-4699-8a8e-288e0d5f68a0)

mia-guo-ux:
**Select**
![Selector](https://github.com/ansys/ansys-sphinx-theme/assets/123198029/a43a1a53-8070-4406-b668-1775b556ec75)

mia-guo-ux:
**Dropdown**
![Dropdown](https://github.com/ansys/ansys-sphinx-theme/assets/123198029/ce7fd89f-096c-47fb-84d7-84a8f1b52d2d)

mia-guo-ux:
**Table**
![Table](https://github.com/ansys/ansys-sphinx-theme/assets/123198029/14085dfe-cce8-45fc-aad5-805fc090378d)

mia-guo-ux:
**Scrollbar**
![Scrollbar](https://github.com/ansys/ansys-sphinx-theme/assets/123198029/2fc33f7d-524f-4c08-969d-bc6183144c99)

mia-guo-ux:
**Breadcrumb**
[...]

PR #424, opened 2024-07-23
fix: admonitions style

Revathyvenugopal162:
Associated with https://github.com/ansys/ansys-sphinx-theme/issues/403#issuecomment-2201293227
# fix admonitions 

## light theme
Screenshot_29-7-2024_174256_

## dark theme
Screenshot_29-7-2024_174249_


MaxJPRey:
Nice work @Revathyvenugopal162 .
Thanks for all the presentation related to it yesterday.

jorgepiloto:
I bet it is tricky to get the admonition icon inline with the rest of the text for multi-line admonitions.

If we want to keep the icon next to the title of the admonition, then let's increase the margin between the icon and the left border of the admonition. Otherwise, icons for one-line admonitions and multi-line ones show a margin difference.

Revathyvenugopal162:
> I bet it is tricky to get the admonition icon inline with the rest of the text for multi-line admonitions.
> 
> If we want to keep the icon next to the title of the admonition, then let's increase the margin between the icon and the left border of the admonition. Otherwise, icons for one-line admonitions and multi-line ones show a margin difference.

updated the margin and the screenshots

jorgepiloto:
@Revathyvenugopal162, I can see both the admonition title and the content are left justified and aligned. However, the icon still shows a bigger margin than in the multi-line admonitions than in the multi-line admonitions than in the one-line ones.

Revathyvenugopal162:
> @Revathyvenugopal162, I can see both the admonition title and the content are left justified and aligned. However, the icon still shows a bigger margin than in the multi-line admonitions than in the multi-line admonitions than in the one-line ones.

margin left for both the icons are same, but the right margin after the icons were different  , i just aligned it back, let me know if this works, otherwise i can revert with the small margin in between. 
image

jorgepiloto (approved):

PR #428, opened 2024-07-25
feat: add quarto cheat sheet extension with cheat sheet option

Revathyvenugopal162:
- Update the cheatsheet option
- Adapt to new quarto cheatsheet format

### DIFF
--- a/doc/source/user-guide/options.rst
+++ b/doc/source/user-guide/options.rst
@@ -208,8 +208,8 @@ that contain these keys, in the order given:
    include the directory name relative to the root of the documentation. For example, if the cheat sheet
    is in the ``getting_started`` directory, the file name is ``getting_started/cheat_sheet.qmd``.
 #. ``title``: Title of the cheat sheet to be displayed in the left navigation pane.
-#. ``output_dir``: Directory where the cheat sheet is saved relative to the output directory after building the documentation.
-   By default, the cheat sheet is saved in the ``_build/html/_static/`` directory.
+#. ``pages``: List of names for the pages to include the cheat sheet on. If no value is provided,
+   the cheat sheet is displayed only on the main ``index.html`` file.
 
 Here is an example of how to add the ``cheatsheet`` dictionary to the `html_theme_options`` dictionary:
 
@@ -220,7 +220,6 @@ Here is an example of how to add the ``cheatsheet`` dictionary to the `html_them
             "cheatsheet": {
                 "file": "<file name including the extension of the cheat sheet>",
                 "pages": "<list of names for the pages to include the cheat sheet on>",  # Optional
-                output_dir: "<directory where the cheat sheet is saved relative to the output directory>",  # Optional
             },
         },
     )
--- a/doc/source/user-guide/options.rst
+++ b/doc/source/user-guide/options.rst
@@ -240,6 +240,6 @@ main ``index.rst`` file and the ``learning.rst`` file in its "Getting started" s
 
 .. note::
 
-    To use this feature, you must have the `quarto <https://quarto.org/>` package installed. To create thumbnails of generated pdf files,
+    To use this feature, you must have the `quarto <https://quarto.org/>` package installed. To create thumbnails of generated PDF files,
     the theme is using `pdf2image`. So you should have the ``poppler`` package installed in your system.
     For more information, see the `pdf2image documentation <https://pypi.org/project/pdf2image/>`_.
--- a/doc/source/user-guide/options.rst
+++ b/doc/source/user-guide/options.rst
@@ -198,21 +198,18 @@ your documentation.
 Cheat sheets
 ------------
 
-If a cheat sheet has been created for your PyAnsys library, you can show a thumbnail
-image of it in the left navigation pane of one or more sections of your documentation.
+If a cheat sheet has been created for your PyAnsys library, with ``quarto``, you can
+add it to the left navigation pane of your documentation.
 
 In the ``html_theme_options`` dictionary, you add a child dictionary named ``cheatsheet``
 that contain these keys, in the order given:
 
-#. ``url``: URL of the cheat sheet.
-#. ``title``: Title of the cheat sheet.
-#. ``thumbnail``: URL of the thumbnail image for the cheat sheet.
-#. ``needs_download``: Whether to download the cheat sheet locally during the documentation build.
-   The default is ``False``, in which case the cheat sheet is accessed directly from the provided URL.
-   If ``True``, the cheat sheet is downloaded to the ``_build/html/_static/`` directory and accessed
-   from here.
-#. ``pages``: List of names for the reStructuredText (RST) files to show the cheat sheet on. If no list
-   is provided, the cheat sheet is shown in the left navigation pane of the main ``index.rst`` file.
+#. ``file``: File name including the extension of the cheat sheet. If the file is inside a directory,
+   include the directory name relative to the root of the documentation. For example, if the cheat sheet
+   is in the ``getting_started`` directory, the file name is ``getting_started/cheat_sheet.qmd``.
+#. ``title``: Title of the cheat sheet to be displayed in the left navigation pane.
+#. ``output_dir``: Directory where the cheat sheet is saved relative to the output directory after building the documentation.
+   By default, the cheat sheet is saved in the ``_build/html/_static/`` directory.
 
 Here is an example of how to add the ``cheatsheet`` dictionary to the `html_theme_options`` dictionary:
 
@@ -221,11 +218,9 @@ Here is an example of how to add the ``cheatsheet`` dictionary to the `html_them
     html_theme_options = (
         {
             "cheatsheet": {
-                "url": "<URL of the cheat sheet>",
-                "title": "<title of the cheat sheet>",
-                "thumbnail": "<URL of the thumbnail image>",
-                "needs_download": True,  # True if you want to download and access the cheat sheet locally
+                "file": "<file name including the extension of the cheat sheet>",
                 "pages": "<list of names for the pages to include the cheat sheet on>",  # Optional
+                output_dir: "<directory where the cheat sheet is saved relative to the output directory>",  # Optional
             },
         },
     )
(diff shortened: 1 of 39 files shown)


src/ansys_sphinx_theme/__init__.py
```diff
@@ -446,7 +448,11 @@ def configure_theme_logo(app: Sphinx):
     if not theme_options.get("logo"):
         theme_options["logo"] = pyansys_logo
 
-    if not isinstance(theme_options.get("logo"), str):
+    if isinstance(theme_options.get("logo"), str) and theme_options.get("logo") not in [
+        "ansys",
+        "pyansys",
+        "no_logo",
+    ]:
```
jorgepiloto, line 455: Shouldn't this be part of a separate PR?

src/ansys_sphinx_theme/__init__.py
```diff
@@ -464,6 +470,105 @@ def configure_theme_logo(app: Sphinx):
...
+        Path to the output directory.
+    output_png : str
+        Name of the output PNG file.
+    """
+    try:
+        doc = pymupdf.open(pdf_path)  # open document
+        for page in doc:
+            pix = page.get_pixmap()
+            pix.save(os.path.join(output_dir, output_png))
+    except Exception as e:
```
jorgepiloto, line 491: Let's avoid capturing generic exceptions. Can we narrow which exceptions are involved here?

src/ansys_sphinx_theme/__init__.py
Revathyvenugopal162: > Shouldn't this be part of a separate PR?

yes, when i was testing this PR with pymechanical, it was blocking the build. i will revert this change and open another PR

src/ansys_sphinx_theme/__init__.py
Revathyvenugopal162: Sure, adding that.

Revathyvenugopal162: added in [1544c20](https://github.com/ansys/ansys-sphinx-theme/pull/428/commits/1544c20529ed0969baca26555a6471a42bda43cd)

Revathyvenugopal162:
Testing with https://github.com/ansys/pymechanical/pull/845

src/ansys_sphinx_theme/__init__.py
```diff
@@ -463,6 +378,137 @@ def configure_theme_logo(app: Sphinx):
...
+    file_path = cheatsheet_file.parent
+    output_dir_path = pathlib.Path(app.outdir) / output_dir
+    try:
+        # Add the cheatsheet to the Quarto project
+        result = subprocess.run(
+            ["quarto", "add", "ansys/pyansys-quarto-cheatsheet@v1", "--no-prompt"],
+            cwd=file_path,
+            capture_output=True,
+            text=True,
+        )
```
jorgepiloto, line 461: Let's isolate the version within the scope of this function:

```suggestion
        VERSION = "v1"
        result = subprocess.run(
            ["quarto", "add", f"ansys/pyansys-quarto-cheatsheet@{VERSION}", "--no-prompt"],
            cwd=file_path,
            capture_output=True,
            text=True,
        )
```

src/ansys_sphinx_theme/__init__.py
```diff
@@ -463,6 +378,137 @@ def configure_theme_logo(app: Sphinx):
...
+            file_path = cheatsheet_file.parent / file
+            if file_path.exists():
+                file_path.unlink()
+
+            # if static folder is clean, delete it
+        if not list(cheatsheet_file.parent.glob("_static/*")):
+            cheatsheet_file.parent.joinpath("_static").rmdir()
+
+    except subprocess.CalledProcessError as e:
+        raise RuntimeError(f"Failed to build Quarto cheatsheet: {e}, ensure Quarto is installed.")
```
jorgepiloto, line 503:
```suggestion
    except subprocess.CalledProcessError as e:
        raise RuntimeError(f"Failed to build Quarto cheatsheet: {e}. Ensure Quarto is installed.")
```

jorgepiloto (changes requested):

src/ansys_sphinx_theme/__init__.py
```diff
@@ -464,6 +470,105 @@ def configure_theme_logo(app: Sphinx):
...
+    app : sphinx.application.Sphinx
+        Application instance for rendering the documentation.
+    """
+    cheatsheet_options = app.config.html_theme_options.get("cheatsheet", {})
+
+    if not cheatsheet_options:
+        return
+
+    cheatsheet_file = cheatsheet_options.get("file", "")
+    output_dir = cheatsheet_options.get("output_dir", "")
```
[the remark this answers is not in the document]
Revathyvenugopal162: removed in [82cb447](https://github.com/ansys/ansys-sphinx-theme/pull/428/commits/82cb4473ce6bf1179be5a487690b351e0a2c8953)

src/ansys_sphinx_theme/__init__.py
```diff
@@ -463,6 +378,138 @@ def configure_theme_logo(app: Sphinx):
...
+    output_dir : pathlib.Path
+        Path to the output directory.
+    output_png : str
+        Name of the output PNG file.
+    """
+    try:
+        doc = pymupdf.open(pdf_path)  # open document
+        for page in doc:
+            pix = page.get_pixmap()
+            pix.save(os.path.join(output_dir, output_png))
```
jorgepiloto, line 398: Please, avoid using `os.path`. Use `pathlib` instead:

```suggestion
            pix.save(output_dir / output_png)
```

src/ansys_sphinx_theme/__init__.py
```diff
@@ -463,6 +378,138 @@ def configure_theme_logo(app: Sphinx):
...
+            "_static/ansys.png",
+        ]
+        for file in supplementary_files:
+            file_path = cheatsheet_file.parent / file
+            if file_path.exists():
+                file_path.unlink()
+
+            # if static folder is clean, delete it
+        if not list(cheatsheet_file.parent.glob("_static/*")):
+            cheatsheet_file.parent.joinpath("_static").rmdir()
```
jorgepiloto, line 501:
```suggestion
        # If static folder is clean, delete it
        if not list(cheatsheet_file.parent.glob("_static/*")):
            cheatsheet_file.parent.joinpath("_static").rmdir()
```

jorgepiloto (changes requested):
Almost there. Some minor suggestions and we will merge this.

src/ansys_sphinx_theme/__init__.py
```diff
@@ -24,10 +24,11 @@
 import logging
 import os
 import pathlib
+import subprocess
 from typing import Any, Dict
 
 from docutils.nodes import document
-import requests
+import pymupdf
```
RobPasMue, line 31: Have we verified that we can use this pacakge? It's license is GNU AFFERO GPL 3.0. I'm guessing there is no problem but I am also interested in its approval for other libraries.

RobPasMue (changes requested):

src/ansys_sphinx_theme/__init__.py
RobPasMue: https://www.gnu.org/licenses/agpl-3.0.html

src/ansys_sphinx_theme/__init__.py
```diff
@@ -463,6 +378,138 @@ def configure_theme_logo(app: Sphinx):
...
+    if not cheatsheet_file:
+        return
+
+    cheatsheet_file = pathlib.Path(app.srcdir) / cheatsheet_file
+    file_name = str(cheatsheet_file.name)
+    file_path = cheatsheet_file.parent
+    output_dir_path = pathlib.Path(app.outdir) / output_dir
+    try:
+        # Add the cheatsheet to the Quarto project
+        VERSION = "v1"
```
SMoraisAnsys, line 456: I would move that at the beginning of the file and rename it, e.g. `CHEAT_SHEET_VERSION`

Revathyvenugopal162: sure, moving it

src/ansys_sphinx_theme/__init__.py
jorgepiloto: I do not think this is an issue for other libraries since they use the theme for development. However, it is an issue within this project for building the wheelhouse.

RobPasMue: Yep - that was my concern. Redistributing is not well seen with AGPL 3.0 from waht I understood, and the wheelhouse is a clear redistribution of software...

Revathyvenugopal162: Thank you for raising the issue, I initially used this approach to avoid adding extra dependencies on ImageMagick or Poppler. However, given the issues we encountered, I reverted to using pdf2image and added a note explaining the change. Please review

doc/source/user-guide/options.rst
```diff
@@ -238,23 +232,14 @@ main ``index.rst`` file and the ``learning.rst`` file in its "Getting started" s
...
-    If you set ``"needs_download": True", you should provide the ``html_static_path`` option
-    in the ``conf.py`` file to specify the location for downloading the cheat sheet to.
-
-    For example, if you want to download the cheat sheet to the ``_build/html/_static/`` directory,
-    you should add the following line in the ``conf.py`` file:
-
-    .. code-block:: pycon
-
-        html_static_path = ["_static"]
+    To use this feature, you must have the `quarto <https://quarto.org/>` package installed. To create thumbnails of generated pdf files,
```
Revathyvenugopal162, line 243:
```suggestion
    To use this feature, you must have the `quarto <https://quarto.org/>` package installed. To create thumbnails of generated PDF files,
```

src/ansys_sphinx_theme/__init__.py
```diff
@@ -395,12 +395,12 @@ def convert_pdf_to_png(pdf_path: pathlib.Path, output_dir: pathlib.Path, output_
         Name of the output PNG file.
     """
     try:
-        doc = pymupdf.open(pdf_path)  # open document
-        for page in doc:
-            pix = page.get_pixmap()
-            pix.save(output_dir / output_png)
-    except RuntimeError as e:
-        raise RuntimeError(f"Failed to convert PDF to PNG: {e}")
+        images = convert_from_path(pdf_path, 500)
```
RobPasMue, line 398: Can we do the import in the case we are interested on calling this function only? I am expecting that, since system dependencies are required, this might lead to failure on resolving imports even and not only when calling it. Maybe I'm wrong but I would constraint the import call to this function only

RobPasMue (changes requested):

src/ansys_sphinx_theme/__init__.py
Revathyvenugopal162: yes,i will add that

Revathyvenugopal162:
tested with https://github.com/ansys/pymechanical/pull/845

PR #446, opened 2024-08-08
fix: stable docs indexing package name

Revathyvenugopal162:
Resolve the syntax error in package name fix for https://github.com/ansys/ansys-sphinx-theme/actions/runs/10299970659/job/28508754310

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -10,7 +10,7 @@ on:
       - main
 env:
   MAIN_PYTHON_VERSION : '3.10'
-  PACKAGE_NAME: 'ansys-sphinx-theme'
+  PACKAGE_NAME: 'ansys_sphinx_theme'
   DOCUMENTATION_CNAME: 'sphinxdocs.ansys.com'
   MEILISEARCH_API_KEY: ${{ secrets.MEILISEARCH_API_KEY }}
   MEILISEARCH_PUBLIC_API_KEY: ${{ secrets.MEILISEARCH_PUBLIC_API_KEY }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci_cd.yml
```diff
@@ -177,8 +177,8 @@ jobs:
 
       - name: Scrape the stable documentation to PyMeilisearch
         run: |
-          VERSION=$(python -c "from ${{ env.PACKAGE_NAME }} import __version__; print('.'.join(__version__.split('.')[:2]))")
-          VERSION_MEILI=$(python -c "from ${{ env.PACKAGE_NAME }} import __version__; print('-'.join(__version__.split('.')[:2]))")
+          VERSION=$(python -c "from ansys_sphinx_theme import __version__; print('.'.join(__version__.split('.')[:2]))")
+          VERSION_MEILI=$(python -c "from ansys_sphinx_theme import __version__; print('-'.join(__version__.split('.')[:2]))")
           echo "Calculated VERSION: $VERSION"
```
RobPasMue, line 182: Oh... we changed the namespace?! This is going to have breaking changes for sure on our repos, on the way we import things.

RobPasMue (approved):

.github/workflows/ci_cd.yml
Revathyvenugopal162: > Oh... we changed the namespace?! This is going to have breaking changes for sure on our repos, on the way we import things.

no we didnt change the name space, the ${{ env.PACKAGE_NAME }} is ansys-sphinx-theme which is causing the syntax error

PR #452, opened 2024-08-09
fix: minor style changes

Revathyvenugopal162:
#### before
image
### after
image

## dropdown with large codes, 
fit the content based on the content length, keep 400px as min width

https://github.com/user-attachments/assets/71d82269-a8da-45d7-9ee3-23b635a4b50c

### before
image
### after
image


clatapie (approved):
Thank you for having fixed this @Revathyvenugopal162. It looks awesome!

PR #453, opened 2024-08-09
fix: downgrade the autoapi and keep ``autoapi`` toctree to ``True`` by default

Revathyvenugopal162:
as title says


RobPasMue (approved):
LGTM - thanks @Revathyvenugopal162 - please do a patch release whenever you can with these changes

PR #472, opened 2024-08-21
fix: autoapi extension

### LINKED ISSUE #464: ``[autoapi]`` target and extension does not add the needed extensions by default
See https://github.com/ansys/pyansys-tools-versioning/pull/280

It looks like the extension developed does not include properly the other required extensions ``sphinx-design`` in this specific case. Until I added it manually, the build process kept complaining of not knowing what tab-set is...

It seems that we are not extending the used extensions properly. @Revathyvenugopal162 could you investigate this?

Revathyvenugopal162:
- fix #464 
- Keep the package depth as variable

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -68,10 +68,10 @@ html_theme_options = {
     },
     "ansys_sphinx_theme_autoapi": {
         "project": project,
-        "directory": "src/ansys_sphinx_theme",
+        "directory": "src/ansys_sphinx_theme/examples",
         "output": "examples/api",
         "own_page_level": "function",
-        "package_depth": 2,
+        "package_depth": 1,
     },
     "logo": "ansys",
 }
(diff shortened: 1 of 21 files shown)


doc/source/conf.py
```diff
@@ -93,12 +93,12 @@
...
     "sphinx_copybutton",
-    "sphinx_design",
+    # "sphinx_design",
     "sphinx.ext.intersphinx",
     "sphinx.ext.todo",
     "sphinx_gallery.gen_gallery",
-    "sphinx_jinja",
-    "notfound.extension",
+    # "sphinx_jinja"
+    # "notfound.extension",
```
Revathyvenugopal162, line 101:
```suggestion
    "notfound.extension",
```

RobPasMue (commented):
LGTM thanks @Revathyvenugopal162 - hope this fixes the extension issues

RobPasMue (approved):

Issue #325, opened 2023-11-29
Doc template not using the entire screen


svandenb-dev:
### Description of the modifications

The template does not scale with the screen resolution. This is impacting the rendering. I got complains already from users. Could we improve the space usage on screen ?

repo : https://github.com/ansys-internal/pyansys-edb

![image](https://github.com/ansys/ansys-sphinx-theme/assets/74993647/9442ddd8-400a-44ab-b14a-2759b206d288)

![image](https://github.com/ansys/ansys-sphinx-theme/assets/74993647/7804c492-1632-4e41-b590-1fe01e8cb9a3)

ansbmartin:
I use the following [CSS](https://sphinxdocs.ansys.com/version/stable/user_guide/css.html#ref-user-guide-css) for this:
```css
.bd-main .bd-content .bd-article-container {
   max-width: unset;
}
.bd-page-width {
   max-width: unset;
}
table {
   width: 100%;
}
.bd-sidebar-primary {
   width: 15%;
}
```

PR #473, opened 2024-08-22
feat: add different width for different media for main content

Revathyvenugopal162:
fix https://github.com/ansys/ansys-sphinx-theme/issues/325
## Now
image

## before
image

### DIFF
--- a/src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/css/pydata-sphinx-theme-custom.css
+++ b/src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/css/pydata-sphinx-theme-custom.css
@@ -730,6 +730,12 @@ button.btn.version-switcher__button:hover {
 
 @media (min-width: 960px) {
   .bd-page-width {
-    max-width: min(100%, 1600px) !important;
+    max-width: min(100%, 1700px) !important;
+  }
+}
+
+@media (min-width: 960px) {
+  .bd-main .bd-content .bd-article-container {
+    width: min(100%, 1200px);
   }
 }
--- a/src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/css/pydata-sphinx-theme-custom.css
+++ b/src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/css/pydata-sphinx-theme-custom.css
@@ -14,7 +14,12 @@
   max-height: calc(100vh - var(--pst-header-height) - 1vh);
   line-height: var(--ast-global-line-height);
   border: none;
-  overflow: hidden;
+  overflow: auto;
+}
+
+.bd-sidebar-primary .sidebar-primary-items__end {
+  margin-bottom: unset;
+  margin-top: unset;
 }
 
 .sidebar-secondary-item {
--- a/src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/css/pydata-sphinx-theme-custom.css
+++ b/src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/css/pydata-sphinx-theme-custom.css
@@ -741,6 +741,6 @@ button.btn.version-switcher__button:hover {
 
 @media (min-width: 960px) {
   .bd-main .bd-content .bd-article-container {
-    width: min(100%, 1200px);
+    width: min(100%, 1070px);
   }
 }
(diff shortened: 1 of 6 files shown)


RobPasMue (approved):
Looks good! I wouldn't be greedier though like @SMoraisAnsys mentioned 😄 you might start finding issues if you stick too much to the left side... 😄

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/css/pydata-sphinx-theme-custom.css
```diff
@@ -715,8 +715,21 @@ button.btn.version-switcher__button:hover {
...
+
+@media (min-width: 960px) {
+  .col-lg-3 {
+    width: fit-content;
+  }
+}
+
+@media (min-width: 960px) {
+  .bd-page-width {
+    max-width: min(100%, 1600px) !important;
```
Revathyvenugopal162, line 733: @SMoraisAnsys, @RobPasMue increase this size to 1700px will increase the screen size, since 1920 is the standard resolution , i can increase this if you prefer (if you are greedy)

Revathyvenugopal162: image

SMoraisAnsys: The result is great ! Is there a reason for the middle part to be that far away from the scroll bar ?
As far as I like this, I don't have enough hindsight to assess the extent to which this could cause problems (which @RobPasMue  pointed out).

Revathyvenugopal162: > The result is great ! Is there a reason for the middle part to be that far away from the scroll bar ? As far as I like this, I don't have enough hindsight to assess the extent to which this could cause problems (which @RobPasMue pointed out).

good catch, i reduced the gap
image

germa89: I think that is going to far.... you probably dont want to have so much space between el central content and the right bar, and so little space between the central content and the scrollbar.....

germa89 (approved):
I like the new change... but:

- dont go to far with it, it will clutter the view.
- Dont stick too close to any of the sides while you have plenty of space on the other side. You want to strick a balance between both sides.

But overall, I like it. Great job!

RobPasMue:
> * Don't go too far with it, it will clutter the view.
> * Don't stick too close to any of the sides while you have plenty of space on the other side. You want to strick a balance between both sides.


Balance is the key!

germa89:
asdf

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/css/pydata-sphinx-theme-custom.css
Revathyvenugopal162: The left and right content is having the same margin and padding, and the scroll bar appears only if there is an overflow in the content. 
image
image
These are the 2 changes with and without scrollbar. pinging @mia-guo-ux also here for insights

mia-guo-ux: Thanks for pinning me. When the text exceeds the container, I suggest truncating the text with "..." and show the full content upon hovering instead of enabling horizontal scrolling. The vertical scrollbar looks good to me. Also...I agree with the comments that balance is the key :D

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/css/pydata-sphinx-theme-custom.css
```diff
@@ -720,8 +720,27 @@ button.btn.version-switcher__button:hover {
...
+
+@media (min-width: 960px) {
+  .bd-page-width {
+    max-width: min(100%, 1700px) !important;
+  }
+}
+
+@media (min-width: 960px) {
+  .bd-main .bd-content .bd-article-container {
+    width: min(100%, 1200px);
```
Revathyvenugopal162, line 744:
```suggestion
    width: min(100%, 1070px);
```

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/css/pydata-sphinx-theme-custom.css
Revathyvenugopal162: Thank you all, Then i will go with the above specifications, and about horizontal scrollbar, i will open another PR to fix.

PR #495, opened 2024-09-09
feat: add member_order to autoapi

klmcadams:
The members are ordered by source rather than alphabetically, so this change should allow people to organize their documentation in alphabetical order if they want

autoapi_member_order: https://sphinx-autoapi.readthedocs.io/en/latest/reference/config.html#confval-autoapi_member_order


RobPasMue (approved):
LGTM @klmcadams - thanks for the impl!

PR #509, opened 2024-09-24
fix: change file location for `search.json` file

Revathyvenugopal162:


src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/js/fuse_search.js
```diff
@@ -80,10 +80,20 @@ require(["fuse"], function (Fuse) {
...
 
+  const pathParts = window.location.pathname.split("/");
+  let basePath;
+
+  // Check if the path contains a version directory
+  if (pathParts.includes("version")) {
+    basePath = pathParts.slice(0, 3).join("/"); // Extract up to /version/{version}
+  } else {
+    basePath = ""; // If there's no version, just use the base path
+  }
```
jorgepiloto, line 91: This is tricky. What if a library has a folder named "version" with some RST files inside?

Revathyvenugopal162: yes, rewrite the code and now we are getting the relative path to search.json file using {{pathto}} jinja in html and the relative path to the root using data-content root from html header

jorgepiloto (approved):
Thanks!

PR #515, opened 2024-09-27
fix: add title breadcrumbs

Revathyvenugopal162:
image


src/ansys_sphinx_theme/search/fuse_search.py
```diff
@@ -45,10 +45,21 @@ def __init__(self, doc_name, app):
...
         self._doc_name = doc_name
         self.doc_path = f"{self._doc_name}.html"
+        self.env = app.env
         self.doc_title = app.env.titles[self._doc_name].astext()
         self._doc_tree = app.env.get_doctree(self._doc_name)
         self.sections = []
 
+    def get_title_breadcrumbs(self):
+        """Generate title breadcrumbs from the document name."""
+        docs_parts = self._doc_name.split("/")[:-1]
```
jorgepiloto, line 55: Could we document what is the last element and why are we excluding it? Just so we do not forget.

src/ansys_sphinx_theme/search/fuse_search.py
```diff
@@ -67,14 +78,31 @@ def iterate_through_docs(self):
...
+            )
+
     @property
     def indices(self):
         """Get search index."""
         for sections in self.sections:
+            self.get_title_breadcrumbs()
+            title_breadcrumbs = self.construct_title_breadcrumbs(sections["section_title"])
+
             search_index = {
```
jorgepiloto, line 102:
```suggestion
            yield {
```

src/ansys_sphinx_theme/search/fuse_search.py
```diff
@@ -45,10 +45,21 @@ def __init__(self, doc_name, app):
         """
         self._doc_name = doc_name
         self.doc_path = f"{self._doc_name}.html"
+        self.env = app.env
         self.doc_title = app.env.titles[self._doc_name].astext()
         self._doc_tree = app.env.get_doctree(self._doc_name)
         self.sections = []
 
+    def get_title_breadcrumbs(self):
```
jorgepiloto, line 53: What is the difference between this method and `construct_title_breadcrumbs`?

jorgepiloto (commented):
Some questions to give a better feedback on this.

src/ansys_sphinx_theme/search/fuse_search.py
jorgepiloto: No need to have this variable if we are just yielding it.

src/ansys_sphinx_theme/search/fuse_search.py
```diff
@@ -67,18 +68,60 @@ def iterate_through_docs(self):
                 }
             )
 
+    def construct_title_breadcrumbs(self, section_title: str) -> str:
```
jorgepiloto, line 71: Let's use the singular form for breadcrumb. This is also [provided by Bootstrap](https://getbootstrap.com/docs/4.0/components/breadcrumb/).

```suggestion
    def construct_title_breadcrumbs(self, section_title: str) -> str:
```

src/ansys_sphinx_theme/search/fuse_search.py
```diff
@@ -67,18 +68,60 @@ def iterate_through_docs(self):
...
+            if self.section_title == self.doc_title:
+                return f"{self.breadcrumbs_title} > {section_title}"
+            else:
+                return f"{self.breadcrumbs_title} > {self.doc_title} > {section_title}"
+        else:
+            return (
+                section_title
+                if self.section_title == self.doc_title
+                else f"{self.doc_title} > {section_title}"
+            )
```
jorgepiloto, line 111: Since there is no code below this `else`, we can remove it and reduce indentation by one level:

```suggestion
        return (
            section_title
            if self.section_title == self.doc_title
            else f"{self.doc_title} > {section_title}"
        )
```

src/ansys_sphinx_theme/search/fuse_search.py
```diff
@@ -67,18 +68,60 @@ def iterate_through_docs(self):
...
+        title_string = [
+            self.env.titles[doc_part].astext()
+            for doc_part in docs_parts
+            if doc_part in self.env.titles
+        ]
+
+        self.breadcrumbs_title = " > ".join(title_string)
+
+        if self.breadcrumbs_title:
+            if self.section_title == self.doc_title:
```
jorgepiloto, line 102: I see what you are doing here. This is nice! However, sometimes, conditions obfuscate what we are checking for. You can isolate the condition, which allows for naming and reusing it later:

```
has_duplicated_items = self.section_title == self.doc_title
```

jorgepiloto: Also, the logic below can be compressed to:

```suggestion
    # Join the titles with " > "
    breadcrumbs = " > ".join(title_parts)
    
    # Check whether to avoid including duplicated items
    ignore_doc_title = self.section_title == self.doc_title

    # Determine the final breadcrumb output
    if ignore_doc_title:
        breadcrumb = f"{breadcrumbs} > {section_title}" if breadcrumbs else section_title
    else:
        breadcrumb = f"{breadcrumbs} > {self.doc_title} > {section_title}" if breadcrumbs else f"{self.doc_title} > {section_title}"
    
    return breadcrumb
```

jorgepiloto (changes requested):
The logic is working as expected. My comments go in the line of naming and compressing the conditions to make those more readable.

src/ansys_sphinx_theme/search/fuse_search.py
```diff
@@ -29,84 +29,124 @@
 
 
 class SearchIndex:
-    """Class to get search index."""
+    """Generates a search index for a Sphinx document."""
```
jorgepiloto, line 32: Verb must be imperative in docstrings:

```suggestion
    """Generate a search index for a Sphinx document."""
```

src/ansys_sphinx_theme/search/fuse_search.py
```diff
@@ -29,84 +29,124 @@
...
+            The section title to generate breadcrumbs for.
+
+        Returns
+        -------
+        str
+            Breadcrumb string.
+
+        Notes
+        -----
+        The last part of `doc_name` is ignored because it represents the file name,
```
jorgepiloto, line 86: Thanks!

jorgepiloto (approved):
Thanks incorporating all the feedback. Approving this. Feel free to merge, @Revathyvenugopal162.

PR #541, opened 2024-10-10
fix: add more components to the search indexing

### LINKED ISSUE #540: Improve indexing of content
Currently the static search is not working that great whenever our API reference is based on autoapi.

See example: 

image

I know for a fact that there is a method called "create_design", that is available at https://geometry.docs.pyansys.com/version/stable/api/ansys/geometry/core/modeler/Modeler.html#Modeler.create_design

However, the static search is not indexing properly subsections inside my API page - only summary, description and some other elements are being indexed, making the search completely broken. We should fix this because currently the static search is not usable for our API reference.

Pinging @ansys/pyansys-core for visibility

Revathyvenugopal162:
fix #540


src/ansys_sphinx_theme/search/fuse_search.py
```diff
@@ -55,9 +55,18 @@ def build_sections(self):
             section_title = node[0].astext()
             section_text = "\n".join(
                 n.astext()
-                for node_type in [nodes.paragraph, nodes.literal_block]
+                for node_type in [
```
jorgepiloto, line 58: We need to find a better solution for this. I think projects should be able to index on a page-basis. If we allow globbing, it could be very powerful, allowing users to customize the index level on chapter-basis too.

```
index_patterns = {
    "api*": ALL_NODES,
    "examples*": PARAGRAPH,
    ...
}
```

Revathyvenugopal162: yeah sure, we can try to get this

RobPasMue: works for me

doc/source/conf.py
```diff
@@ -71,6 +71,11 @@
         "threshold": 0.2,
         "limit": 7,
         "minMatchCharLength": 3,
+        "ignoreLocation": True,
+        "index_patterns": {
+            "examples/api/*": "ALL_NODES",
+            "examples/sphinx_examples/*": "TITLES + PARAGRAPHS",
```
jorgepiloto, line 77: These shouldn't be strings but aliases implemented in our `__init__.py` file.

```python
PARAGRAPHS = [docutils.nodes.paragraph]
```

Thus, when users use them, we just get a long list of docutils nodes:

```python
TILES + PARAGRAPHS
```

src/ansys_sphinx_theme/__init__.py
```diff
@@ -28,13 +28,20 @@
...
 
-from docutils.nodes import document
+from docutils import nodes
 from sphinx import addnodes
 from sphinx.application import Sphinx
 
 from ansys_sphinx_theme.extension.linkcode import DOMAIN_KEYS, sphinx_linkcode_resolve
 from ansys_sphinx_theme.latex import generate_404  # noqa: F401
-from ansys_sphinx_theme.search import create_search_index, update_search_config
+from ansys_sphinx_theme.search import (  # noqa: F401
```
jorgepiloto, line 37: We should use the `__all__` to avoid ignoring these errors.

src/ansys_sphinx_theme/search/__init__.py
```diff
@@ -38,6 +44,14 @@ def update_search_config(app: Sphinx) -> None:
     theme_static_options["keys"] = ["title", "text"]
     theme_static_options["threshold"] = theme_static_options.get("threshold", 0.5)
     theme_static_options["limit"] = theme_static_options.get("limit", 10)
+    app.add_config_value("index_patterns", {}, "html")
```
jorgepiloto, line 47: Why adding the configuration here? Shouldn't we read it from the `app.config`?

src/ansys_sphinx_theme/search/fuse_search.py
```diff
@@ -45,19 +58,21 @@ def __init__(self, doc_name, app):
         self.doc_name = doc_name
         self.doc_path = f"{self.doc_name}.html"
         self.env = app.env
+        self.theme_options = app.config.html_theme_options.get("static_search", {})
```
jorgepiloto, line 61: This is confusing. Why naming it `theme_options` if you are retrieving the `static_search` dict?

src/ansys_sphinx_theme/search/fuse_search.py
```diff
@@ -126,6 +141,20 @@ def _title_to_anchor(title: str) -> str:
     return re.sub(r"[^\w\s-]", "", title.lower().strip().replace(" ", "-"))
 
 
+def get_pattern_for_each_page(app, doc_name):
+    """Get the pattern for each page in the search index."""
+    if getattr(app.env.config, "index_patterns"):
+        patterns = app.env.config.index_patterns
+    else:
+        patterns = {}
```
jorgepiloto, line 149: What about?

```suggestion
    patterns = app.env.config.get("index_patterns", {})
```

src/ansys_sphinx_theme/search/fuse_search.py
```diff
@@ -27,11 +27,24 @@
...
+LITERAL = [nodes.literal]
+ALL_NODES = [nodes.Text]
+ALL_NODES_WITHOUT_RAW = [  # type: ignore
+    nodes.paragraph,
+    nodes.title,
+    nodes.list_item,
+    nodes.field_list,
+    nodes.compound,
+    nodes.block_quote,
+]
```
jorgepiloto, line 41: This is strange. The `ALL_NODES` contains only one element whereas the `ALL_NODES_WITHOUT_RAW` has even more. Why is that?

jorgepiloto (changes requested):
Looking good. Just added some comments and suggestions, @Revathyvenugopal162.

src/ansys_sphinx_theme/search/fuse_search.py
Revathyvenugopal162: I was just testing some node parameters. Yes, the 'all node' refers to all the raw text inside the docs, and it also includes the JS, as I mentioned privately. This was just for testing purposes. I'll clean it up a bit and incorporate your comments as well.

src/ansys_sphinx_theme/search/fuse_search.py
```diff
@@ -143,7 +143,7 @@ def _title_to_anchor(title: str) -> str:
 
 def get_pattern_for_each_page(app, doc_name):
     """Get the pattern for each page in the search index."""
-    patterns = app.env.config.get("index_patterns", {})
+    patterns = app.env.config.index_patterns or {}
```
jorgepiloto, line 146: Interesting... I though that the `app.env.config` was a dictionary and the method `get` could be used.

doc/source/conf.py
```diff
@@ -172,7 +181,8 @@ def extract_example_links(
     list
         List of example links.
     """
-    g = Github()
+    token = os.getenv("GITHUB_TOKEN")
+    g = Github(token)
```
Revathyvenugopal162, line 185:
```suggestion
    g = Github()
```

Revathyvenugopal162:
testing it in https://github.com/ansys/pyansys-geometry/actions/runs/11459949702

jorgepiloto:
Testing the latest changes...

jorgepiloto:
Nice! This is working.

src/ansys_sphinx_theme/search/fuse_search.py
```diff
@@ -68,6 +87,31 @@ def build_sections(self):
...
+                    if element_child.tagname != "desc_signature":
+                        continue
+                    section_anchor_id = element_child.attributes["ids"][0]
+            section_text = element.astext()
+            section_title = f"{title} > {section_anchor_id}"
+            self.sections.append(
+                {
+                    "title": section_title,
+                    "text": section_text,
+                    "anchor_id": section_anchor_id,
```
jorgepiloto, line 111: This is working fine. However, the only anchors containing a "." are the ones generated by the Python domain. Even if users create a `The CONTRIBUTING.md anchor`, the "." gets replaced automatically by a "-".

Knowing this, would it be possible to split the anchor using as character the "."? If so, then we should keep the latest item, which is the actual method or property.

This leads to a nice breadcrumb like `FooClass > Methods > foo_method` instead of `FooClass > Methods > FooClass.foo_method`.

Revathyvenugopal162: i agree, let me try to get that.

Revathyvenugopal162: [89081f3](https://github.com/ansys/ansys-sphinx-theme/pull/541/commits/89081f394fe644e0055c455e5a34ebf6360883ac) added this

doc/source/conf.py
```diff
@@ -133,6 +142,8 @@
     rst_epilog += f.read()
 
 
+linkcheck_exclude_documents = ["changelog"]
```
jorgepiloto, line 145: Why is this required?

src/ansys_sphinx_theme/search/fuse_search.py
```diff
@@ -68,6 +87,31 @@ def build_sections(self):
                 }
             )
 
+            self._process_desc_element(node, section_title)
+
+    def _process_desc_element(self, node, title):
+        """Process `desc` element within a section."""
```
jorgepiloto, line 93: Could we clarify about what is this `disc` object?

doc/source/user-guide/options.rst
```diff
@@ -124,6 +124,23 @@ the ``delay`` key in the ``static_search`` dictionary with a value specifying
...
+Here is an example of how to add the ``index_patterns`` dictionary to the `conf.py`` file:
+
+.. code-block:: python
+
+    index_patterns = {
+        "api/": ALL_NODES,
+        "examples/": TITLES + PARAGRAPHS,
+    }
+
+The above example will index all nodes in the ``api/`` directory and only titles and paragraphs in the ``examples/`` directory.
```
jorgepiloto, line 141:
```suggestion
The above example indexes all nodes in the ``api/`` directory and only titles and paragraphs in the ``examples/`` directory.
```

jorgepiloto (changes requested):
Just left some minor comments. Almost ready to be merged.

src/ansys_sphinx_theme/search/fuse_search.py
Revathyvenugopal162: The `desc` object in Sphinx is a **node used for documenting items** like functions, classes, and other domain-specific objects. This node isn’t directly added by `docutils.addnodes`; rather, it’s defined in [**Sphinx’s own `sphinx.addnodes` module**](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/addnodes.py).

To index the properties and functions in API references, we need to use the desc node.

doc/source/conf.py
Revathyvenugopal162: I noticed a 502 error for link checks in the changelog file—likely a temporary issue due to server overload https://github.com/ansys/ansys-sphinx-theme/actions/runs/11496793507/job/31999155151 , I ignored it for the moment, but I checked, and it’s no longer an issue, so I’ll revert the changes.

doc/source/conf.py
```diff
@@ -62,9 +65,15 @@
         "threshold": 0.2,
         "limit": 7,
         "minMatchCharLength": 3,
+        "ignoreLocation": True,
```
Revathyvenugopal162, line 68:
```suggestion
```

doc/source/conf.py
```diff
@@ -133,6 +142,8 @@
     rst_epilog += f.read()
 
 
+# linkcheck_exclude_documents = ["changelog"]
+
```
Revathyvenugopal162, line 146:
```suggestion
```

jorgepiloto (approved):

jorgepiloto:
Thanks for adding this, @Revathyvenugopal162!

PR #542, opened 2024-10-10
fix: add default `enter` for search

### LINKED ISSUE #539: Search bar on entry should default to a search page (ideally PyData's search)
Whenever you perform a search like the following one:

image

And you press enter (i.e. you do not select any option) - you are always redirected to the first result of the static search engine. We should avoid doing this and if the user presses enter/clicks "search" we should take them to a search page with all the available entries. In fact, until we have refined our static search algorithm, I think we should default to PyData's search engine IMO because currently the static search algo is not working properly for API reference for example.

Pinging @ansys/pyansys-core  for visibility

Revathyvenugopal162:
fix #539 

1. Pressing Enter without typing anything in the search bar will take you to the PyData search page.
2. Pressing Enter on any of the entries in the suggestion bar will take you to the corresponding anchor.
3. After typing a search query, if the suggestions aren't enough, pressing Enter will take you to the PyData search page.

### DIFF
--- a/doc/changelog.d/542.fixed.md
+++ b/doc/changelog.d/542.fixed.md
@@ -1,0 +1,1 @@
+fix: add default `enter` for search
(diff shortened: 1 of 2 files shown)


RobPasMue (approved):
LGTM

Issue #526, opened 2024-10-03
Allow developers to skip the build of the examples


jorgepiloto:
Every time the documentation of this project is built, various requests get performed to fetch third party examples.

Multiple runs lead to these third parties blocking for 5 min any additional requests.

We should allow to skip the build of the examples via the Jinja approach we have in other projects.

### Note

Before contributing and starting any work, please make sure you agree with the [CLA](https://developer.ansys.com/form/cla-acceptance) and sign it in your PR.

Revathyvenugopal162:
**Steps to Implement:**

- [ ]  Ensure that `sphinx_jinja` is included as a dependency in pyproject.toml (it is already present).
- [ ] Implement Build Example Environment Variable: https://github.com/ansys/pyansys-geometry/blob/main/doc/source/conf.py#L353
- [ ] Update `examples.rst`: https://github.com/ansys/pyansys-geometry/blob/main/doc/source/index.rst?plain=1#L41
**Acceptance Criteria:**

- [ ] - The build process can successfully skip examples based on the environment variable.
- [ ] - The Jinja template is properly integrated and functioning in `index.rst`.
- [ ] - All changes are tested, and documentation is updated accordingly.

PR #553, opened 2024-10-17
doc: allow developer to skip examples build

jorgepiloto:
This pull-request fixes https://github.com/ansys/ansys-sphinx-theme/issues/526 when the ``BUILD_EXAMPLES=true`` environment variable is set.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -253,10 +253,10 @@ else:
     }
 
     nbsphinx_prolog = """
-    Download this example as a :download:`Jupyter notebook </{{ env.docname }}.ipynb>`.
+Download this example as a :download:`Jupyter notebook </{{ env.docname }}.ipynb>`.
 
-    ----
-    """
+----
+"""
     nbsphinx_execute = "always"
     nbsphinx_thumbnails = {
         "examples/nbsphinx/jupyter-notebook": "_static/pyansys_light_square.png",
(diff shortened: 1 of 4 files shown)


doc/source/conf.py
```diff
@@ -243,26 +253,33 @@ def download_and_process_files(example_links: List[str]) -> List[str]:
...
+    )
+
+    admonitions_links = download_and_process_files(admonitions_links)
+    todo_include_todos = True  # admonition todo needs this to be True
+
+    jinja_contexts["examples"] = {"inputs_examples": file_names}
+    jinja_contexts["admonitions"] = {"inputs_admonitions": admonitions_links}
+
+
+print(f"JINJA CONTEXTS: {jinja_contexts}")
```
Revathyvenugopal162, line 285:
```suggestion
```
do we need to print this?

Revathyvenugopal162 (approved):

doc/source/conf.py
jorgepiloto: Good catch! Fixing it.

jorgepiloto:
I can't reproduce locally the error raised by the CI/CD pipelines...

SMoraisAnsys (commented):
@jorgepiloto I think the error is related to the last section title, it should be "## Render a data frame" instead of "# Render a data frame".

jorgepiloto:
> @jorgepiloto I think the error is related to the last section title, it should be "## Render a data frame" instead of "# Render a data frame".

Thanks, @SMoraisAnsys. This was not the point but you guided me in the right direction. The problem was with the multi-string for the prologue. It is fixed in https://github.com/ansys/ansys-sphinx-theme/pull/553/commits/1fc76c9ee59f93e6d213138d4b6d71d9e118b347

PR #554, opened 2024-10-17
fix: padding for toctree entry in sidebar

### LINKED ISSUE #543: TOC tree of API moves a bit to the right when clicked
When you click on an element of the ToC tree of the API reference, it moves a bit to the left, making it look like it's the children of the element on top.

[Screen Recording - Made with FlexClip.webm](https://github.com/user-attachments/assets/373c456a-ae40-4bc5-90a3-c4165dd6ea70)

Revathyvenugopal162:
fix #543 
## before
image

## After
image

### DIFF
--- a/doc/changelog.d/554.fixed.md
+++ b/doc/changelog.d/554.fixed.md
@@ -1,0 +1,1 @@
+fix: padding for toctree entry in sidebar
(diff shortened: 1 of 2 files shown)


jorgepiloto (approved):
Tested locally. All working as expected. Thanks, @Revathyvenugopal162.

PR #557, opened 2024-10-18
fix: add the default search options in the config env

Revathyvenugopal162:


RobPasMue (approved):

SMoraisAnsys (approved):
LGTM, thanks for fixing this !
Just wondering, would it make sense to log the static options such that one could be aware of such options from the log in CICD ?

jorgepiloto:
> LGTM, thanks for fixing this ! Just wondering, would it make sense to log the static options such that one could be aware of such options from the log in CICD ?

So far, we are logging some of the events in the web browser console. Messages on our side start with the prefix `[AST]`, the acronym for Ansys Sphinx Theme.

PR #565, opened 2024-10-23
fix: ensure app.builder.outdir is a Path object

jorgepiloto:


src/ansys_sphinx_theme/search/fuse_search.py
```diff
@@ -147,6 +148,6 @@ def create_search_index(app, exception):
         search_index.build_sections()
         search_index_list.extend(search_index.indices)
 
-    search_index_path = app.builder.outdir / "_static" / "search.json"
+    search_index_path = Path(app.builder.outdir.resolve().absolute()) / "_static" / "search.json"
```
RobPasMue, line 151: AFAIK - -absolute returns a Path already
```suggestion
    search_index_path = app.builder.outdir.resolve().absolute() / "_static" / "search.json"
```

src/ansys_sphinx_theme/search/fuse_search.py
```diff
@@ -23,6 +23,7 @@
 """Module for generating search indices."""
 
 import json
+from pathlib import Path
```
RobPasMue, line 26:
```suggestion
```
Consequently

src/ansys_sphinx_theme/search/fuse_search.py
jorgepiloto, line 151:
```suggestion
    search_index_path = Path(app.builder.outdir) / "_static" / "search.json"
```

jorgepiloto: Good point! I just realized that can't apply any of the Path methods because `app.builder.outdir` is a `str` on PyDPF side. Thus, I am casting it to a Path before applying the any of the `resolve()`, or `absolute()` methods.

jorgepiloto:
Finally working in https://github.com/ansys/pydpf-core/actions/runs/11477549649/job/31942226295?pr=1787

jorgepiloto:
Once merged, I will cherry-pick this into v1.1 and perform a patch so PyDPF can update.

PR #572, opened 2024-10-29
fix: exclude search files

### LINKED ISSUE #570: Allow certain files to be excluded from static search

Revathyvenugopal162:
fix #570


src/ansys_sphinx_theme/search/fuse_search.py
```diff
@@ -204,7 +204,17 @@ def create_search_index(app, exception):
...
 
-    for document in app.env.found_docs:
+    static_search_options = app.config.html_theme_options.get("static_search", {})
+    excluded_docs = static_search_options.get("files_to_exclude", [])
+
+    for docs in excluded_docs:
+        included_docs = [
+            doc
+            for doc in app.env.found_docs
+            if not (doc.startswith(docs) if docs.endswith("/") else doc == docs)
```
jorgepiloto, line 214: Could we refactor this line? The logic is tricky here. If we could use boolean variables explaining the idea that would be great.

Revathyvenugopal162: sure, added

jorgepiloto (approved):

doc/source/user-guide/options.rst
```diff
@@ -124,6 +124,29 @@ the ``delay`` key in the ``static_search`` dictionary with a value specifying
 the amount of milliseconds to wait before executing the search. A value of
 ``0`` disables the debounce function.
 
+Additionally, you can decide the limit of the search results by setting the
+``limit`` key in the ``static_search`` dictionary. The default value is ``10``.
+
+To exclude some directories from the search index, you can use the
```
SMoraisAnsys, line 130:
```suggestion
To exclude files or directories from the search index, you can use the
```

src/ansys_sphinx_theme/search/fuse_search.py
```diff
@@ -204,7 +204,26 @@ def create_search_index(app, exception):
 
     search_index_list = []
 
-    for document in app.env.found_docs:
+    static_search_options = app.config.html_theme_options.get("static_search", {})
+    excluded_docs = static_search_options.get("files_to_exclude", [])
+
+    for exclude_doc in excluded_docs:
+        is_folder = exclude_doc.endswith("/")
```
SMoraisAnsys, line 211: Wouldn't using `os.path.isdir(document)` be better ? I'm fearing that the ending with `"/"` might be OS dependent

src/ansys_sphinx_theme/search/fuse_search.py
```diff
@@ -204,7 +204,26 @@ def create_search_index(app, exception):
...
+        # Filter out documents based on whether exclude_doc is a folder or a file:
+        # - If a folder, exclude all documents starting with the folder name.
+        # - If a file, exclude only the exact file name.
+        included_docs = [
+            doc
+            for doc in app.env.found_docs
+            if not (is_folder and doc.startswith(exclude_doc)) and doc != exclude_doc
+        ]
+
+    for document in included_docs:
```
SMoraisAnsys, line 226: Shouldn't you define a default value for `included_docs` ? If `excluded_docs` is empty, you'll never define that variables.

SMoraisAnsys (commented):
Great feature ! I left some minor comments :)

src/ansys_sphinx_theme/search/fuse_search.py
jorgepiloto: Yes, that would be the best, @Revathyvenugopal162. Thanks for pointing this out, @SMoraisAnsys. Does Sphinx treat these as actual paths, Revathy? If so, they may even have an `isdir()` method.

src/ansys_sphinx_theme/search/fuse_search.py
Revathyvenugopal162: added default value, thanks Sebastian

src/ansys_sphinx_theme/search/fuse_search.py
Revathyvenugopal162: > Yes, that would be the best, @Revathyvenugopal162. Thanks for pointing this out, @SMoraisAnsys. Does Sphinx treat these as actual paths, Revathy? If so, they may even have an `isdir()` method.

No, it is just the docnames, but we can try with output directory

src/ansys_sphinx_theme/search/fuse_search.py
```diff
@@ -204,7 +204,26 @@ def create_search_index(app, exception):
...
-    for document in app.env.found_docs:
+    static_search_options = app.config.html_theme_options.get("static_search", {})
+    excluded_docs = static_search_options.get("files_to_exclude", [])
+    included_docs = app.env.found_docs
+    output_dir = Path(app.builder.outdir)
+
+    for exclude_doc in excluded_docs:
+        exclude_doc = Path(exclude_doc)
+        exclude_path = output_dir / exclude_doc
+        is_folder = exclude_path.is_dir()
```
Revathyvenugopal162, line 215: @jorgepiloto @SMoraisAnsys added the path , please review

jorgepiloto: I would play with the `relative_to` method. Something like this:

```python
    for path in all_paths:
        resolved_path = Path(path).resolve()

        # Check if the resolved path is not in the ignore set and not a subdirectory of any ignored path
        if not any(resolved_path.is_relative_to(ignore_path) for ignore_path in ignore_set):
            yield path
```

src/ansys_sphinx_theme/search/fuse_search.py
```diff
@@ -204,7 +204,26 @@ def create_search_index(app, exception):
...
+
+        # Exclude documents based on whether exclude_doc is a folder or a file:
+        # - For folders, exclude all documents within the folder.
+        # - For files, exclude only the exact file match.
+        included_docs = [
+            doc
+            for doc in included_docs
+            if not (is_folder and Path(doc).parent == exclude_doc) and Path(doc) != exclude_doc
```
SMoraisAnsys, line 223: @Revathyvenugopal162 Sorry but I'm not sure about the following LOC.
Let's say that you exlcude `my_folder`, doesn't that mean that `my_folder/subfolder/file` won't be ignored since it isn't the direct parent ?
Could we use `parents` instead to retrieve avoid that problem ?

Also, would it make sense to compute included_docs as follow:
```python
excluded_docs = static_search_options.get("files_to_exclude", [])
excluded_docs = [Path(document).resolve() for document in excluded_docs]
filtered_docs = []
for document in app.env.found_docs:
    document_path = Path(document).resolve()
    if document_path in excluded_docs:
        continue
    if any(parent.resolve() in excluded_docs for parent in document_path.parents):
        continue
    filtered_docs.append(document)
```

SMoraisAnsys (commented):
I feel like there is still a caveat here. I left a comment

src/ansys_sphinx_theme/search/fuse_search.py
Revathyvenugopal162: @SMoraisAnsys For the moment, i implemented exlcude ``my_folder``, will exclude ``my_folder/subfolder/file`` as well. You wanted the other way ?

SMoraisAnsys:
I wasn't aware of `is_relative_to`, I like the final result. Also I double checked and `p.is_relative_to(p)` returns `True` so everything is fine ! The result is very elegant, well done !

PR #580, opened 2024-11-06
Enable 'show_prev_next' in the documented defaults

greschd:
Switch the default value of `show_prev_next` to True in the theme options documentation.

The reason for this change is that the prev / next buttons are useful for navigating the documentation, especially in narrow viewports.
In that case, the only other option to navigate is the hamburger menu, which the user may not be aware of.

### DIFF
--- a/doc/changelog.d/580.documentation.md
+++ b/doc/changelog.d/580.documentation.md
@@ -1,0 +1,1 @@
+Enable 'show_prev_next' in the documented defaults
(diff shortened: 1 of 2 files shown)


greschd:
> In that case, the only other option to navigate is the hamburger menu, which the user may not be aware of.

The change is motivated by a user who ended up unaware that the documentation even _had_ more than a single page (for a doc which contained the README as first page; this was a while ago).

SMoraisAnsys (approved):
I'm not a big fan of the rendering, but since it's only associated to the options documentation and can help a user, then all the better.

Revathyvenugopal162 (approved):

PR #583, opened 2024-11-07
feat: add whatsnew options

Revathyvenugopal162:
- [x] Add `whatsnew` html and theme options
- [x] Add changelog template for global use
- [x] add css style file


pyproject.toml
```diff
@@ -118,15 +121,16 @@ show_missing = true
 
 [tool.towncrier]
 directory = "doc/changelog.d"
-filename = "doc/source/changelog.rst"
-template = "doc/changelog.d/changelog_template.jinja"
+filename = "doc/source/release-note.rst"
+template = "ansys_sphinx_theme:changelog_template.jinja"
 start_string = ".. towncrier release notes start\n"
-title_format = "`{version} <https://github.com/ansys/ansys-sphinx-theme/releases/tag/v{version}>`_ - {project_date}"
+title_format = "Version `{version} <https://github.com/ansys/pymechanical/releases/tag/v{version}>`_ ({project_date})"
```
Revathyvenugopal162, line 127:
```suggestion
title_format = "Version `{version} <https://github.com/ansys/ansys-sphinx-theme/releases/tag/v{version}>`_ ({project_date})"
```

.github/workflows/ci_cd.yml
```diff
@@ -45,6 +45,19 @@ jobs:
...
+        with:
+          python-version: ${{ env.MAIN_PYTHON_VERSION }}
+
+      - name: "Checkout repo"
+        uses: actions/checkout@v4
+
+      - name: "Install dependencies"
+        run: |
+          python -m pip install --upgrade pip
+          pip install -e .[changelog]
```
jorgepiloto, line 59: I see... These are required for Sphinx design, right?

.github/workflows/ci_cd.yml
```diff
@@ -45,6 +45,19 @@ jobs:
       pull-requests: write
     runs-on: ubuntu-latest
     steps:
+      - name: "Setup python"
+        uses: actions/setup-python@v2
```
jorgepiloto, line 49: Outdated version. Latest is:

```suggestion
        uses: actions/setup-python@v5
```

doc/source/changelog.rst
```diff
@@ -1,8 +1,19 @@
...
 #############
 
+
+.. button-ref:: release-note
+    :ref-type: doc
+    :color: primary
+    :shadow:
+    :expand:
+
+    Go to latest release notes
```
jorgepiloto, line 15: I am not a big fan of duplicating files. I understand we have changed the format. However, why having a `CHANGELOG` and a `RELEASE NOTES` page? For me, it is confusing.

src/ansys_sphinx_theme/__init__.py
```diff
@@ -538,6 +538,83 @@ def check_for_depreciated_theme_options(app: Sphinx):
...
+def extract_whatsnew(app, doctree, docname):
+    """Extract the what's new content from the document."""
+    config_options = app.config.html_theme_options
+
+    whats_new_options = config_options.get("whatsnew")
+    if not whats_new_options:
+        return
+    no_of_contents = whats_new_options.get("no_of_headers", 3)
+    document_name = whats_new_options.get("file", "release-note")
+    get_doctree = app.env.get_doctree(document_name)
```
jorgepiloto, line 550: Fixing variable name:

```suggestion
    doctree = app.env.get_doctree(document_name)
```

src/ansys_sphinx_theme/__init__.py
```diff
@@ -538,6 +538,83 @@ def check_for_depreciated_theme_options(app: Sphinx):
...
+    whats_new_content = []
+    docs_content = get_doctree.traverse(nodes.section)
```
jorgepiloto, line 552:
```suggestion
    docs_content = doctree.traverse(nodes.section)
```

src/ansys_sphinx_theme/__init__.py
```diff
@@ -538,6 +538,83 @@ def check_for_depreciated_theme_options(app: Sphinx):
...
+    """Extract the what's new content from the document."""
+    config_options = app.config.html_theme_options
+
+    whats_new_options = config_options.get("whatsnew")
+    if not whats_new_options:
+        return
+    no_of_contents = whats_new_options.get("no_of_headers", 3)
+    document_name = whats_new_options.get("file", "release-note")
+    get_doctree = app.env.get_doctree(document_name)
+    whats_new_content = []
```
jorgepiloto, line 551: Agree on a naming convention. Either `whats_new` or `whatsnew`. Also, these are used later. If possible, instantiate the variables right before using them. If they are used for storing the values originated in a loop, then instantiate it right before the loop.

src/ansys_sphinx_theme/__init__.py
```diff
@@ -538,6 +538,83 @@ def check_for_depreciated_theme_options(app: Sphinx):
...
+        whats_new_nodes = [node for node in sections if node[0].astext().lower() == "whatsnew"]
+
+        if not whats_new_nodes:
+            continue
+
+        children = [node for node in whats_new_nodes[0].traverse(nodes.section)]
+
+        headers = [child[0].astext() for child in children]
+
+        # header_anchors = [child.get("ids")[0] for child in children]
```
jorgepiloto, line 579: Unused comment:

```suggestion
```

src/ansys_sphinx_theme/__init__.py
```diff
@@ -538,6 +538,83 @@ def check_for_depreciated_theme_options(app: Sphinx):
...
+
+        if len(children) > 1:
+            children = headers[1:]
+
+        else:
+            children = [whats_new_nodes[0].traverse(nodes.paragraph)[0].astext()]
```
jorgepiloto, line 585:
```suggestion
        if len(children) > 1:
            children = headers[1:]
        else:
            children = [whats_new_nodes[0].traverse(nodes.paragraph)[0].astext()]
```

src/ansys_sphinx_theme/__init__.py
```diff
@@ -538,6 +538,83 @@ def check_for_depreciated_theme_options(app: Sphinx):
...
+
+
+def add_whatsnew(app, pagename, templatename, context, doctree):
+    """Add what's new section to the context."""
+    config_options = app.config.html_theme_options
+    whats_new_options = config_options.get("whatsnew")
+    if not whats_new_options:
+        return
+
+    pages = whats_new_options.get("pages", ["index"])
```
jorgepiloto, line 606: Does this mean that any page named `index` will display the "What's new" section?

src/ansys_sphinx_theme/templates/changelog_template.jinja
```diff
@@ -0,0 +1,36 @@
+{% if sections["Whatsnew"] %}
+  {%+ for category, val in sections["Whatsnew"].items() +%}
+    {%+ if definitions.get(category) and definitions[category].get('name') +%}
+{{ definitions[category]['name'] }}
+{{ "-" * definitions[category]['name']|length }}
+    {% endif %}
+    {%- for text, values in sections["Whatsnew"][category].items() %}
```
jorgepiloto, line 7: Are you using the `values` variable in here?

jorgepiloto (changes requested):
This is looking really good. Just left some comments.

.github/workflows/ci_cd.yml
Revathyvenugopal162: yes, Incase if there is no sphinx design in their doc dependency

src/ansys_sphinx_theme/__init__.py
Revathyvenugopal162: No only the landing page will have the name `index` all other index pages will be `dir/index` for example `getting-started/index`

src/ansys_sphinx_theme/templates/changelog_template.jinja
Revathyvenugopal162: I can add it if we want to add/remove or we can also give it as an option to the user

doc/source/changelog.rst
RobPasMue: We should only have one page IMO...

Revathyvenugopal162: > I am not a big fan of duplicating files. I understand we have changed the format. However, why having a `CHANGELOG` and a `RELEASE NOTES` page? For me, it is confusing.
I will update the template and revert the files. Currently, there is a title-level inconsistency with the existing template. I intended to have 'What's New' as a section, but doing so creates a title-level inconsistency since the previous template treats its children as sub-subsections. That is the reason i kept the new template in new file.

src/ansys_sphinx_theme/templates/changelog_template.jinja
jorgepiloto: Is it just the issue number or the link to the issue? Can we create a link with it?

Revathyvenugopal162: yes, it is issue number with link. we can add that values also in the template

src/ansys_sphinx_theme/__init__.py
jorgepiloto: Ah! Nice. Thanks for clarifying.

RobPasMue:
@klmcadams - please resolve existing conflicts. It was at a pretty good state, let's try to push it across the line =)

RobPasMue:
@klmcadams @Revathyvenugopal162 - let us all know whenever the PR is ready for review. Marking it as draft mode in the meantime to avoid confusion. I see that at least @klmcadams is still working on it

doc/source/user-guide/options.rst
```diff
@@ -256,3 +256,91 @@ main ``index.rst`` file and the ``learning.rst`` file in its "Getting started" s
...
+
+- Code blocks: Use the following format to include code blocks:
+
+  .. code-block:: rst
+
+        .. code:: python
+
+           print("hello world")
+
+If a format is used in the "content" field that does not fall into the category above, it will not
```
RobPasMue, line 301:
```suggestion
If a format is used in the "content" field that does not fall into the categories above, it will not
```

src/ansys_sphinx_theme/__init__.py
RobPasMue, line 1: Are there actually any differences? GitHub is rendering it as if the entire file was modified... it's hard to review this way :S

klmcadams: There are some new functions - get_whatsnew_options, add_whatsnew_changelog, extract_whatsnew, and add_whatsnew_sidebar. I think the line endings changed in the file. Is there a way to revert the line endings?

doc/source/conf.py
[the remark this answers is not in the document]
klmcadams: I think the line endings made it seem like the entire file was changed

klmcadams:
@RobPasMue I fixed the line endings in `conf.py` & `__init__.py`. My previous commits were in CRLF but the repo files were in LF. I added the .gitattributes file to hopefully prevent this from happening in the future

src/ansys_sphinx_theme/__init__.py
RobPasMue: Much better now, thanks!

src/ansys_sphinx_theme/__init__.py
```diff
@@ -540,6 +552,556 @@ def check_for_depreciated_theme_options(app: Sphinx):
...
+    -------
+    tuple
+        Tuple containing the paths to the whatsnew file, changelog file, and the sidebar pages.
+        If the whatsnew options are not found, return None for each of those fields.
+    """
+    # Get the html_theme_options from conf.py
+    config_options = app.config.html_theme_options
+
+    # Get the whatsnew key from the html_theme_options
+    whatsnew_options = config_options.get("whatsnew", None)
```
RobPasMue, line 573: No need for the default ``None`` - None is already the default.

RobPasMue: Applies everywhere

RobPasMue (approved):
LGTM! As long as the logic works I'm fine with it. Some nice string handling you are doing in this PR @klmcadams

PR #588, opened 2024-11-22
Fix/remove broken references

Andy-Grigg:
I decided to create a draft PR with some possible clean-up or improvements to the theme. I'm not an expert in this area though, and I don't know a lot of the history, so I don't know for sure if these are good changes or not. They're all broadly focused on trying to reduce the number of errors in the browser console. Not all of these only occur when opening the files directly, some occur even when hosted through github, see here https://jobqueue.grantami.docs.pyansys.com/version/dev/index.html.

Summary of the changes I've hacked together so far:

* Remove some `@import` statements from CSS files. As far as I can tell, these files are all imported separately in the HTML. The thing that got me looking in this area is the `@import "../sg_gallery.css";` statement, which fails if the sphinx-gallery extension isn't used
* Use a theme config entry to control whether `ast-search-button.html` is included in the document. If fuse-search isn't being used, this avoids running the javascript which causes errors when not running through a server. Also, the fuse-search.css doesn't seem to exist either.
  * This might be an alternative way of handling the fuse search vs sphinx search thing. I tried this out quickly with the v1.2.0 version of search.js and it sort of works, but on click the results disappear. Probably wouldn't be too much effort if that's the direction you wanted to go.
* Warn about meilisearch, instead of raising a warning. This causes an exception when building, when I think we actually want to be warning instead.

There are still lots of CORS errors when searching, which is strange because they don't occur when using the pydata sphinx theme directly. Some more digging might be able to fix this.

Given that a lot of this relates to using the theme for a closed-source project that should work when opening the files not via a web server, I'm happy to do the legwork in resolving these issues. I just want to make sure it's moving in a direction that everyone's comfortable with before I get too far down the line.

### DIFF
--- a/src/ansys_sphinx_theme/__init__.py
+++ b/src/ansys_sphinx_theme/__init__.py
@@ -585,7 +585,9 @@ def setup(app: Sphinx) -> Dict:
     # Add default HTML configuration
     setup_default_html_theme_options(app)
 
-    update_search_config(app)
+    use_pyansys_search = app.config.html_theme_options.get("use_ansys_search", False)
+    if use_pyansys_search:
+        update_search_config(app)
 
     # Verify that the main CSS file exists
     if not CSS_PATH.exists():
@@ -604,7 +606,8 @@ def setup(app: Sphinx) -> Dict:
     app.connect("html-page-context", add_cheat_sheet)
     app.connect("html-page-context", add_search_option)
     app.connect("build-finished", replace_html_tag)
-    app.connect("build-finished", create_search_index)
+    if use_pyansys_search:
+        app.connect("build-finished", create_search_index)
     return {
         "version": __version__,
         "parallel_read_safe": True,
--- a/src/ansys_sphinx_theme/__init__.py
+++ b/src/ansys_sphinx_theme/__init__.py
@@ -27,6 +27,7 @@ import os
 import pathlib
 import subprocess
 from typing import Any, Dict
+import warnings
 
 from docutils import nodes
 from sphinx import addnodes
@@ -433,6 +434,28 @@ def add_cheat_sheet(
         context["sidebars"] = sidebar
 
 
+def add_search_option(
+    app: Sphinx, pagename: str, templatename: str, context: Dict[str, Any], doctree: nodes.document
+) -> None:
+    """Control the type of searching used.
+
+    Parameters
+    ----------
+    app : ~sphinx.application.Sphinx
+        Application instance for rendering the documentation.
+    pagename : str
+        Name of the current page.
+    templatename : str
+        Name of the template being used.
+    context : dict
+        Context dictionary for the page.
+    doctree : ~docutils.nodes.document
+        The doctree.
+    """
+    use_pyansys_search = app.config.html_theme_options.get("use_ansys_search", False)
+    context["use_ansys_search"] = use_pyansys_search
+
+
 def build_quarto_cheatsheet(app: Sphinx):
     """
     Build the Quarto cheatsheet.
@@ -533,8 +556,10 @@ def check_for_depreciated_theme_options(app: Sphinx):
     """
     theme_options = app.config.html_theme_options
     if "use_meilisearch" in theme_options:
-        raise DeprecationWarning(
-            "The 'use_meilisearch' option is deprecated. Remove the option from your configuration file."  # noqa: E501
+        warnings.warn(
+            "The 'use_meilisearch' option is deprecated. Remove the option "
+            "from your configuration file.",
+            DeprecationWarning,
         )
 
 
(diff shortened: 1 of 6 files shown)


Andy-Grigg:
I should add as well, I am happy to split this up into separate PRs and discuss more in issues if that's more appropriate. Just wanted to capture the result of my hacking on a Friday afternoon.

src/ansys_sphinx_theme/__init__.py
```diff
@@ -433,6 +434,28 @@ def add_cheat_sheet(
...
+        Name of the current page.
+    templatename : str
+        Name of the template being used.
+    context : dict
+        Context dictionary for the page.
+    doctree : ~docutils.nodes.document
+        The doctree.
+    """
+    use_pyansys_search = app.config.html_theme_options.get("use_ansys_search", False)
+    context["use_ansys_search"] = use_pyansys_search
```
jorgepiloto, line 456: Fixing the naming here:

```suggestion
    use_ansys_search = app.config.html_theme_options.get("use_ansys_search", False)
    context["use_ansys_search"] = use_ansys_search
```

src/ansys_sphinx_theme/__init__.py
```diff
@@ -577,8 +604,10 @@ def setup(app: Sphinx) -> Dict:
     app.connect("html-page-context", update_footer_theme)
     app.connect("html-page-context", fix_edit_html_page_context)
     app.connect("html-page-context", add_cheat_sheet)
+    app.connect("html-page-context", add_search_option)
     app.connect("build-finished", replace_html_tag)
-    app.connect("build-finished", create_search_index)
+    if use_pyansys_search:
```
jorgepiloto, line 609: Fixing this name:

```suggestion
    if use_ansys_search:
```

jorgepiloto (commented):
Thanks for fixing this, @Andy-Grigg. Let me check locally that everything works as expected.

For the CORS policy, there is nothing we can do about it. Certain features require the documentation to be served in an HTTP server.

jorgepiloto:
I am having a deep look at the errors. Some are still present and some others are new.

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/layout.html
```diff
@@ -3,6 +3,12 @@
 {# Append our custom CSS after the bootstrap stylesheet so we can override where necessary #}
 {%- block css %}
   {{ super() }}
+
+  <!-- CSS for DataTables rendering -->
+  <script src="https://code.jquery.com/jquery-3.6.4.min.js"></script>
+  <link href="https://cdn.datatables.net/v/dt/dt-2.2.0/datatables.min.css" rel="stylesheet">
+  <script src="https://cdn.datatables.net/v/dt/dt-2.2.0/datatables.min.js"></script>
```
jorgepiloto, line 10: This solves the JQuery issues for the DataTables. However, we are injecting a lot of CSS and JS just for something we barely use. Let me investigate the usage of the `html_context` and see if we can dynamically load this when required.

jorgepiloto:
Most of the errors are solved now. Still working on a jQuery problem...

Andy-Grigg:
> For the CORS policy, there is nothing we can do about it. Certain features require the documentation to be served in an HTTP server.

I'm not worried about the CORS policy at all. It's a well-known security restriction that exists for good reasons and isn't something that experienced users should be surprised to see.

Thank you for picking this up, I really appreciate it! I'll give this branch a test locally and see how I get on.

jorgepiloto:
Hi @Andy-Grigg, all errors and warnings on our side have been fixed. The only ones I get now are for loaded but unused font-awesome CSS files. These are on PyData side.

Andy-Grigg:
I'm still seeing some JS errors related to searching:

image

The first two appear when I click in the search box, and the third appears when I type a character. Each character triggers a new exception. The search still works, opening from a server and as a file, so it's not a blocker.

jorgepiloto:
> @jorgepiloto Also, these changes make it required to define the following in conf.py:
> 
> ```python
> html_context = {
>     "page_assets": {},
> }
> ```
> 
> Is it possible to have an empty dict added in the theme if one isn't specified, before the processing begins?

That's a good point! Let me fix it.

jorgepiloto:
> @jorgepiloto I'm still seeing some JS errors related to searching:
> 
> image
> 
> The first two appear when I click in the search box, and the third appears when I type a character. Each character triggers a new exception. The search still works, opening from a server and as a file, so it's not a blocker.

Strange... I can not see these. Did you tried using an incognito tab to ensure no cache is present?

Andy-Grigg:
I see it in incognito mode as well, served up both using http.server and PyCharm's built-in server. I see it in Edge and Firefox.

Andy-Grigg:
Whoops, this was me doing something stupid.

I had some local uncommitted changes left over in my local copy of this branch. Did a git reset and the problems have gone away. Sorry about that!

This is now all good from my end!

jorgepiloto:
Thanks for double checking, @Andy-Grigg.

PR #612, opened 2025-02-04
fix: astroid version

Revathyvenugopal162:


RobPasMue (approved):

pyproject.toml
```diff
@@ -35,6 +35,7 @@ autoapi = [
     "sphinx-autoapi==3.4.0",
     "sphinx-design==0.6.1",
     "sphinx-jinja==2.0.2",
+    "astroid>=3.3.8,<4.0",
```
RobPasMue, line 38: Pretty restrictive though, right? Why not above 3.3 or even 3?

Revathyvenugopal162: sure, added 3.

RobPasMue: Awesome, all good now!

RobPasMue (approved):

MaxJPRey:
Nice catch @Revathyvenugopal162 and @RobPasMue 
Thanks for your responsiveness on this.

PR #630, opened 2025-02-20
fix: typo in autoapi template

Revathyvenugopal162:
Found this typo when there was a change in a library with a nested class using the template, which caused it to fail to render.

The macro was defined as `render_autosummary_template `and used in the function as `autosummary_template`, which caused the issue.


SMoraisAnsys (approved):
LGTM, thanks for patching this issue ! I tested it and it worked :)

MaxJPRey (approved):

PR #635, opened 2025-03-03
feat: add python support

Revathyvenugopal162:


doc/source/conf.py
```diff
@@ -96,7 +96,7 @@
 
 # Intersphinx mapping
 intersphinx_mapping = {
-    "python": ("https://docs.python.org/3", None),
+    "python": ("https://docs.python.org/3.13", None),
```
RobPasMue, line 99:
```suggestion
    "python": ("https://docs.python.org/3", None),
```
I would keep it as ``3`` rather than pointing specifically to ``3.13``.

RobPasMue (approved):

PR #652, opened 2025-03-17
fix: update the style for highlight name function

### LINKED ISSUE #643: Wrong font family in code blocks
Some code blocks in documentation render certain words without monospace font:

https://mapdl.docs.pyansys.com/version/stable/examples/gallery_examples/01-geometry/02-areas.html#area-selection

Image

This is happening in other internal projects that do not override the CSS of this theme. Therefore, we must ensure a particular font family for code-blocks.

Revathyvenugopal162:
fix #643 

| light | dark |
| - | - |
| Screenshot 2025-03-24 at 09 15 45 | Screenshot 2025-03-24 at 09 16 37 |

## update 

code fontstyle to `**Source Code Pro", monospace**`

Screenshot 2025-03-24 at 09 11 30


clatapie:
Out of curiosity, I noticed a different color is used for ``in`` and ``for``  in dark mode but not in light mode. Also, ``range`` doesn't seem to be highlighted in dark mode but is highlighted in light one. 
Is it common to have different highlights between dark and light modes?

src/ansys_sphinx_theme/assets/styles/pydata-sphinx-theme-custom.scss
```diff
@@ -779,3 +779,12 @@ button.btn.version-switcher__button:hover {
 span.math {
   display: math;
 }
+
+/**
+* code block name function
+*/
+
+div.highlight .nf{
+  font-family: var(--ast-body-family);
```
jorgepiloto, line 788: We need a dedicated monospace font for code. So far, our body font is "Open Sans, sans-serif". Any chance we can have an `--ast-code-family`?

jorgepiloto (changes requested):
Let's ensure we use a dedicated monospace font-family for code blocks.

src/ansys_sphinx_theme/assets/styles/pydata-sphinx-theme-custom.scss
MaxJPRey: That would be great!

Revathyvenugopal162:
> Out of curiosity, I noticed a different color is used for `in` and `for` in dark mode but not in light mode. Also, `range` doesn't seem to be highlighted in dark mode but is highlighted in light one. Is it common to have different highlights between dark and light modes?
Great observation, @clatapie ! The primary reason for the changes is that we use different pigment systems for code blocks in dark and light themes, which naturally leads to variations in keyword (or function) highlighting and distinct color palettes for readability and contrast in each mode. The second issue could be that we applied the changes only to the light theme and not the dark one, which we can fix in our theme. I’ll check this before merging the PR. As for the pigments, we can find the most similar and reasonable ones for both light and dark themes.

src/ansys_sphinx_theme/assets/styles/pydata-sphinx-theme-custom.scss
Revathyvenugopal162: yeah sure . adding the monospace to code

src/ansys_sphinx_theme/assets/styles/pydata-sphinx-theme-custom.scss
```diff
@@ -779,3 +779,17 @@ button.btn.version-switcher__button:hover {
 span.math {
   display: math;
 }
+
+/**
+* code block name function
+*/
```
jorgepiloto, line 785:
```suggestion
/**
 * code block name function
 */
```

jorgepiloto (approved):
Thanks! Just left one suggestion for a docstring.

Issue #649, opened 2025-03-12
Missing parameters in docstring


MaxJPRey:
The parameters description are missing in the docstring.

https://github.com/ansys/ansys-sphinx-theme/blob/4a767bda3dd85ed73f8dc7a6647c4b7d20dd559e/src/ansys_sphinx_theme/__init__.py#L422

PR #655, opened 2025-03-25
fix: change logging info to error in quarto build

### LINKED ISSUE #650: Log an error instead of info
In the following code, I would suggest to log an `error` and not an `info` if possible as the content is from `stderr`.

https://github.com/ansys/ansys-sphinx-theme/blob/4a767bda3dd85ed73f8dc7a6647c4b7d20dd559e/src/ansys_sphinx_theme/__init__.py#L430

Revathyvenugopal162:
fix #650  #649

### DIFF
--- a/src/ansys_sphinx_theme/__init__.py
+++ b/src/ansys_sphinx_theme/__init__.py
@@ -419,7 +419,15 @@ def convert_pdf_to_png(pdf_path: pathlib.Path, output_dir: pathlib.Path, output_
 
 
 def run_quarto_command(command, cwd):
-    """Run quarto command and logs its output."""
+    """Run quarto command and logs its output.
+
+    Parameters
+    ----------
+    command : list
+        List of command arguments.
+    cwd : str
+        Current working directory.
+    """
     command = ["quarto"] + command
     try:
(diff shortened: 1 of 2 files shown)


RobPasMue (approved):

src/ansys_sphinx_theme/__init__.py
```diff
@@ -419,15 +419,23 @@ def convert_pdf_to_png(pdf_path: pathlib.Path, output_dir: pathlib.Path, output_
...
-    """Run quarto command and logs its output."""
+    """Run quarto command and logs its output.
+
+    Parameters
+    ----------
+    command : list
+        List of command arguments.
+    cwd : str
+        Current working directory.
+    """
```
RobPasMue, line 430: If you know the type hints, why not adding them to the method signature?

RobPasMue (changes requested):

src/ansys_sphinx_theme/__init__.py
RobPasMue: My bad - shouldn't have approved that quickly 😄

PR #658, opened 2025-03-26
fix: quarto error logging

Revathyvenugopal162:
Quarto writes both stdout and stderr to stderr so we need to log it as info if it's not an error

error noticed in https://github.com/ansys/pymapdl/actions/runs/14087805194/job/39456752974?pr=2346#step:22:189


jorgepiloto (approved):
Thanks, @Revathyvenugopal162

PR #689, opened 2025-05-09
fix: smaller screen view and search for mobile views

Revathyvenugopal162:
partially fix #683 

- [x] Adjust the width of the search bar to allign with nav bar
- [x] Search is now working with smaller media

> [!NOTE]
>  Open the details to see images in different media 

<details><summary>Details</summary>
<p>

| media | images |
|--------|--------|
| phone | Screenshot 2025-05-12 at 15 08 35 |
| phone 2 | Screenshot 2025-05-12 at 15 09 44 |
| tablet | Screenshot 2025-05-12 at 15 09 57 |
| laptop | Screenshot 2025-05-12 at 10 24 36 | 

</p>
</details> 

<details><summary>More images</summary>
<p>

## Ansys actions with this branch
Screenshot 2025-05-12 at 10 25 04

## Ansys actions docs before

Screenshot 2025-05-12 at 10 12 36

</p>
</details>

PR #691, opened 2025-05-12
feat: small layout adjustments

germa89:
- Center navbar elements
- Centering `bd-content` container. This is more obvious when we hide any of the sidebars.
- Moving search bar to `navbar_end` so it hides when there is no space. This change probably needs some thoughts. Since the search bar "disappear" for small screens and gets located in the primary sidebar:

image

#### Before
image

#### After
image

#### Before
image

#### After
image

#### Before
image

#### After
image

### DIFF
--- a/src/ansys_sphinx_theme/__init__.py
+++ b/src/ansys_sphinx_theme/__init__.py
@@ -156,8 +156,11 @@ def setup_default_html_theme_options(app):
     # Place all switchers and icons at the end of the navigation bar
     if theme_options.get("switcher"):
         theme_options.setdefault(
-            "navbar_end", ["version-switcher", "theme-switcher", "navbar-icon-links"]
+            "navbar_end",
+            ["search-field", "version-switcher", "theme-switcher", "navbar-icon-links"],
         )
+
+    theme_options.setdefault("navbar_persistent", [])
     theme_options.setdefault("collapse_navigation", True)
     theme_options.setdefault("navigation_with_keys", True)
 
--- a/src/ansys_sphinx_theme/__init__.py
+++ b/src/ansys_sphinx_theme/__init__.py
@@ -157,7 +157,7 @@ def setup_default_html_theme_options(app):
     if theme_options.get("switcher"):
         theme_options.setdefault(
             "navbar_end",
-            ["search-field", "version-switcher", "theme-switcher", "navbar-icon-links"],
+            ["search-button-field", "version-switcher", "theme-switcher", "navbar-icon-links"],
         )
 
     theme_options.setdefault("navbar_persistent", [])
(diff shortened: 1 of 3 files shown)


src/ansys_sphinx_theme/__init__.py
```diff
@@ -156,8 +156,11 @@ def setup_default_html_theme_options(app):
     # Place all switchers and icons at the end of the navigation bar
     if theme_options.get("switcher"):
         theme_options.setdefault(
-            "navbar_end", ["version-switcher", "theme-switcher", "navbar-icon-links"]
+            "navbar_end",
+            ["search-field", "version-switcher", "theme-switcher", "navbar-icon-links"],
```
Revathyvenugopal162, line 160:
```suggestion
            ["search-button-field", "version-switcher", "theme-switcher", "navbar-icon-links"],
```

Revathyvenugopal162:
I'm okay with centering the navbar, especially when there is no sidebar. However, if I'm not mistaken, the search might not work properly with this layout. I can test it locally to confirm.

germa89:
> I'm okay with centering the navbar, especially when there is no sidebar. However, if I'm not mistaken, the search might not work properly with this layout. I can test it locally to confirm.

Please do, I haven't really checked the functionality.

**I have updated the PR text**

germa89:
It seems to work???

image

image

Not sure if this is what you meant. Anyway, feel free to test and let me know because I am implementing this changes in PyMAPDL too.

Revathyvenugopal162:
> It seems to work???
> 
> image image
> Not sure if this is what you meant. Anyway, feel free to test and let me know because I am implementing this changes in PyMAPDL too.

This is the search from pydata sphinx theme, not from our theme, Any way adding this change should work https://github.com/ansys/ansys-sphinx-theme/pull/691/files/ca60b135ef165289874da5da489ac1cfb086eede#r2084245571 , Thank you so much for the PR, After we merge this PR i will update #689 also, so that the laptop view will get adjusted

src/ansys_sphinx_theme/__init__.py
germa89: I guess the difference is that `search-button-field` shows the keyboard shortcut?

germa89: > search field.html is the pop-up search bar from pydata sphinx theme, that we avoid completely as per ux instruction, and search button field is the inbuild search for layout. We overwrite the https://github.com/ansys/ansys-sphinx-theme/blob/main/src/ansys_sphinx_theme/theme/ansys_sphinx_theme/components/search-button-field.html search button field in our components, expecially some custom class names. 
>
> so to use ansys sphinx theme, search we might use search-button-field

Agreed!

germa89:
@Revathyvenugopal162 implemented changes!

Revathyvenugopal162 (approved):

PR #704, opened 2025-05-26
feat: add deprecated warning for `ìndex_pattern``search option

Revathyvenugopal162:


src/ansys_sphinx_theme/search/fuse_search.py
```diff
@@ -214,6 +216,17 @@ def create_search_index(app, exception):
     excluded_docs = static_search_options.get("files_to_exclude", [])
     included_docs = app.env.found_docs
 
+    patterns = app.env.config.index_patterns or {}
+    if patterns:
+        # Raise warning that this feature will deprecate in the future
+        logger.info(
```
RobPasMue, line 222: Why not an actual warning?

PR #713, opened 2025-05-28
Update README.rst to fix label from black to ruff

tkoyama010:
I noticed that this repository is using ruff after watching black badge in README.rst.


RobPasMue (approved):
Thank you @tkoyama010!

PR #729, opened 2025-06-11
fix: raise exception for deprecated index pattern

Revathyvenugopal162:


src/ansys_sphinx_theme/search/fuse_search.py
```diff
@@ -193,25 +193,6 @@ def create_search_index(app, exception):
...
-    patterns = app.env.config.index_patterns or {}
-    if patterns:
-        # Raise warning that this feature will deprecate in the future
-        logger.warning(
-            "The 'index_patterns' configuration option is deprecated and will be removed"
-            " in v1.5.0 onwards. "
-            "It will index all paragraphs, titles, and literals by default. "
-            "More information can be found in the documentation at https://sphinxdocs.ansys.com.",
-        )
-
```
RobPasMue, line 214: Can we leave it and raise an error for now? Maybe until 1.6 or even 1.7. Many people ignore the warnings - even if they shouldn't.. and they also ignore the errors sometimes...

I would raise an actual exception for a few minor releases now. And completely remove it later.

src/ansys_sphinx_theme/search/fuse_search.py
```diff
@@ -193,6 +193,17 @@ def create_search_index(app, exception):
...
+
+    # TODO (RV): Deprecate this in release 0.16.0
+    # https://github.com/ansys/ansys-sphinx-theme/issues/730
+    patterns = app.env.config.index_patterns or {}
+    if patterns:
+        logger.error(
+            "The 'index_patterns' is depreciated and no longer supported. "
+            "Please see the documentation https://sphinxdocs.ansys.com for the "
+            "new search configuration options."
+        )
```
RobPasMue:
```suggestion
        error = ValueError(
            "The 'index_patterns' is deprecated and no longer supported. "
            "Please see the documentation https://sphinxdocs.ansys.com for the "
            "new search configuration options."
        )
        logger.error(error)
        raise error
```

Revathyvenugopal162: I am not able to add these suggestion since i pushed some change, may be at the same time you suggested, sorry for that, adapted this in https://github.com/ansys/ansys-sphinx-theme/pull/729/commits/1fedf2894ca5ca5bfd59ef020466a2b37692d5fd

RobPasMue (approved):
Did the suggestion again - let's move to TypeError better. Approving though.

PR #754, opened 2025-07-17
fix: minor improvements in ``jinja`` auto escape and ``subprocess`` call

Revathyvenugopal162:


src/ansys_sphinx_theme/cheatsheet.py
```diff
@@ -27,7 +27,7 @@
 """
 
 import pathlib
-import subprocess
+import subprocess  # nosec: B404
```
RobPasMue, line 30: All nosec statements should be followed with at least an in-line comment justifying this decision right above

Revathyvenugopal162:
> LGTM, could you just explain why the `autoescape` feature?

By default, Jinja2 does not enable auto-escaping, which means HTML content passed through template variables is not automatically escaped unless explicitly configured. Enabling auto-escaping ensures that input strings are safely filtered before rendering.

src/ansys_sphinx_theme/cheatsheet.py
Revathyvenugopal162: Sure, added in https://github.com/ansys/ansys-sphinx-theme/pull/754/commits/8b7f66207866ed5f3ce1e33dccbd67eab077e0ca

PR #758, opened 2025-07-21
feat: add multiple tags for `meta-data`

Revathyvenugopal162:
This pull request introduces functionality to support multiple physics tags in the PyAnsys HTML metadata. The changes include updates to both the documentation and the HTML layout template to handle lists of physics tags.

### DIFF
--- a/doc/source/user-guide/configuration.rst
+++ b/doc/source/user-guide/configuration.rst
@@ -148,7 +148,7 @@ If you want to add multiple physics tags, you can do so by providing a list of t
        "pyansys_tags": ["Electromagnetics", "Structures", "Fluids"],
    }
 
-This will result in multiple `<meta>` tags in your HTML pages:
+This result in multiple `<meta>` tags in your HTML pages:
 
 .. code-block:: html
 
--- a/doc/source/user-guide/configuration.rst
+++ b/doc/source/user-guide/configuration.rst
@@ -140,6 +140,23 @@ which result in the following metadata in your HTML pages:
     <meta property="og:site_name" content="PyAnsys" />
     <meta name="physics" content="Electromagnetics" />
 
+If you want to add multiple physics tags, you can do so by providing a list of tags:
+
+.. code-block:: python
+
+   html_context = {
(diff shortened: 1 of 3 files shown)


doc/source/user-guide/configuration.rst
```diff
@@ -140,6 +140,23 @@ which result in the following metadata in your HTML pages:
...
 
+If you want to add multiple physics tags, you can do so by providing a list of tags:
+
+.. code-block:: python
+
+   html_context = {
+       "pyansys_tag": ["Electromagnetics", "Structures", "Fluids"],
+   }
+
+This will result in multiple `<meta>` tags in your HTML pages:
```
Revathyvenugopal162, line 151:
```suggestion
This result in multiple `<meta>` tags in your HTML pages:
```

doc/source/user-guide/configuration.rst
```diff
@@ -140,6 +140,23 @@ which result in the following metadata in your HTML pages:
     <meta property="og:site_name" content="PyAnsys" />
     <meta name="physics" content="Electromagnetics" />
 
+If you want to add multiple physics tags, you can do so by providing a list of tags:
+
+.. code-block:: python
+
+   html_context = {
+       "pyansys_tag": ["Electromagnetics", "Structures", "Fluids"],
```
jorgepiloto, line 148: If this must be an iterable, then let's rename it to `pyansys_tags`.


```suggestion
       "pyansys_tags": ["Electromagnetics", "Structures", "Fluids"],
```

jorgepiloto (changes requested):
Let one comment about the name of the variable according to the type its holding.

doc/source/user-guide/configuration.rst
Revathyvenugopal162: @jorgepiloto Thank you for review, So you prefer to keep, pyansys_tag and pyansys_tags at same time? because the layout has to change based on that, see https://github.com/ansys/ansys-sphinx-theme/pull/758/files#diff-fe62002009468a01eec84a28b20a06966e119f306c93d39f7aaed250c258ed59

jorgepiloto (approved):

PR #769, opened 2025-08-05
fix: add home section

### LINKED ISSUE #257: Modify the link from the pyansys logo in docs
### Description of the modifications

just wondering if the below makes more sense as general guidance for doc home page navigation? Any thoughts?

image

### Useful links and references

_No response_

### LINKED ISSUE #498: Add home section in the navigation bar
here is the proposed solution by @mia-guo-ux 
image

 - Add <project name>-Home in the nav section
 - Change Home icon to project name

pinging @ansys/pyansys-core  for visibility

### LINKED ISSUE #702: Stop allowing users to customize logo link
As per an internal discussion with the @ansys/documentation, it was agreed to restrict the freedom of users when setting the logo. The reason is to ensure consistency across our projects.

- For those using `html_theme_options = { "logo": "pyansys" }`, we must redirect to <https://docs.pyansys.com>
- for those using `html_theme_options = { "logo": "ansys" }`, we must redirect to <https://ansys.com>

This introduces some breaking changes. However, I do not think this option has been widely adopted by projects.

Revathyvenugopal162:
Fix #702, fix #498, fix #257

Screenshot 2025-08-14 at 09 53 46

Screenshot 2025-08-14 at 09 57 56


src/ansys_sphinx_theme/__init__.py
```diff
@@ -474,7 +477,43 @@ def update_search_sidebar_context(
...
+    for toc in traverse_or_findall(root_toc, toctree):
+        if not toc.attributes.get("entries"):
+            return
+
+        for title, page in toc.attributes["entries"]:
+            if title == "Package Home":
+                return
+
+        home_entry = (
+            nodes.Text("Package Home"),
```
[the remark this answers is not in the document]
Revathyvenugopal162: Mia suggested having the package at home in [this comment](https://github.com/ansys/ansys-sphinx-theme/issues/498#issuecomment-3155648187) last week, but we can discuss this further.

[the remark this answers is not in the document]
Revathyvenugopal162: pinging @mia-guo-ux  here , for insights

src/ansys_sphinx_theme/__init__.py
```diff
@@ -474,7 +478,76 @@ def update_search_sidebar_context(
...
+    docname : str
+        Name of the current document.
+
+    Notes
+    -----
+    This function checks if the 'Home' entry already exists in the root TOC.
+    If it does not exist, it adds the 'Home' entry at the beginning of the TOC.
+    The 'Home' entry links to the index page of the documentation.
+    """
+    index_page = "index"
```
jorgepiloto, line 499: We should retrieve this from the `master_doc` in the `conf.py`.

src/ansys_sphinx_theme/__init__.py
```diff
@@ -474,7 +478,76 @@ def update_search_sidebar_context(
...
+    if exception:
+        return
+
+    outdir = pathlib.Path(app.outdir)
+    project_name = f"{app.config.html_short_title} home" or None
+    if not project_name:
+        project_name = f"{app.config.project} home" or "Package Home"
+
+    for html_file in outdir.rglob("*.html"):
+        with html_file.open("r", encoding="utf-8") as f:
```
jorgepiloto, line 540: Avoid using single letter variables:


```suggestion
        with html_file.open("r", encoding="utf-8") as file:
```

jorgepiloto (changes requested):
Almost there. I just requested some changes.

pyproject.toml
```diff
@@ -30,25 +30,26 @@ classifiers = [
     "Programming Language :: Python :: 3.13",
 ]
 dependencies = [
-    "Sphinx>=6.1.0",
-    "pydata-sphinx-theme>=0.15.4,<0.17",
-    "Jinja2>=3.1.2",
+    "beautifulsoup4==4.13.4",
```
jorgepiloto, line 33: Not a big fan of adding this dependency just for adding a "Home" directory knowing that a simple would do the job...

```rst
Home <self>
```

jorgepiloto (approved):

src/ansys_sphinx_theme/__init__.py
```diff
@@ -474,7 +478,76 @@ def update_search_sidebar_context(
...
+    -------
+    None
+    """
+    if exception:
+        return
+
+    outdir = pathlib.Path(app.outdir)
+    project_name = f"{app.config.html_short_title} home" or None
+    if not project_name:
+        project_name = f"{app.config.project} home" or "Package Home"
```
SMoraisAnsys, line 537: I'm confused with `project_name = f"{app.config.html_short_title} home" or None`, wouldn't that never lead to project_name being equal to None ? Thus, I would change the if statement with "None home" or " home" depending on the kind of expected values.

SMoraisAnsys (changes requested):
Thanks for performing this PR @Revathyvenugopal162 !!! I left a minor comment.

Agreeing with @jorgepiloto's comment, would it be possible to reach the same result without using an extra dependency ?

Revathyvenugopal162:
> Thanks for performing this PR @Revathyvenugopal162 !!! I left a minor comment.
> 
> Agreeing with @jorgepiloto's comment, would it be possible to reach the same result without using an extra dependency ?

yes, i am trying locally without using the package.

src/ansys_sphinx_theme/__init__.py
Revathyvenugopal162: Ha true, i will change that

Revathyvenugopal162: updated

src/ansys_sphinx_theme/__init__.py
```diff
@@ -474,7 +478,80 @@ def update_search_sidebar_context(
...
+    """
+    if exception:
+        return
+
+    outdir = pathlib.Path(app.outdir)
+    if not app.config.html_short_title and not app.config.project:
+        project_name = "Package Home"
+
+    else:
+        project_name = f"{app.config.html_short_title} home" or f"{app.config.project} home"
```
SMoraisAnsys, line 539:
```suggestion
    project_name = "Package Home"
    if app.config.html_short_title:
        project_name = f"{app.config.html_short_title} home"
    elif app.config.project:
        project_name = f"{app.config.project} home"
```
I think this would do what you wanted to

Revathyvenugopal162: added, sorry i couldnt add this by suggestion with github gui, since i added some changes in between

Revathyvenugopal162:
@jorgepiloto @SMoraisAnsys removed the bs4, and added tooltip with regex
Screenshot 2025-08-27 at 10 26 14

src/ansys_sphinx_theme/__init__.py
```diff
@@ -474,7 +481,85 @@ def update_search_sidebar_context(
...
+        project_name = f"{app.config.project} home"
+
+    for html_file in outdir.rglob("*.html"):
+        text = html_file.read_text(encoding="utf-8")
+
+        def replacer(match):
+            attrs_before, attrs_after = match.groups()
+            full_attrs = f"{attrs_before}{attrs_after}"
+            if "title=" in full_attrs:
+                return match.group(0)  # don’t duplicate title
```
SMoraisAnsys, line 553:
```suggestion
                return match.group(0)  # don't duplicate title
```

SMoraisAnsys (approved):
I'm lacking knowledge on the code associated to tooltip addition. However, it seems to make sense.

Issue #532, opened 2024-10-07
Clean up JavaScript files by removing redundant code


Revathyvenugopal162:
**Description:**  
Refactor and optimize existing [JavaScript files](https://github.com/ansys/ansys-sphinx-theme/tree/main/src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/js) by removing redundant code, simplifying logic, and ensuring all event listeners and functions are necessary and efficient. This will improve code readability, reduce file size, and enhance performance.

**Acceptance Criteria:**

- [ ] - Remove any unused functions or variables.
- [ ] - Consolidate duplicate code blocks or functions.
- [ ] - Test all functionality to ensure no regressions are introduced.

KartikSaxena19:
can you assign this task to me

jorgepiloto:
Hi @KartikSaxena19, sure! Remember to sign our [CLA] and add your name to the list of `CONTRIBUTORS`.

[CLA]: https://developer.ansys.com/form/cla-acceptance

KartikSaxena19:
As I'm new to this platform can you please help me that from where i can solve your issue ,how to download ansys if required and where to test the code.

jorgepiloto:
Hi @KartikSaxena19, you do not need to download any of the [Ansys] software for contributing to this project. This is a pure Python library that acts as a documentation theme for [Sphinx]. Thus, it ships some [css files] and [js files].

The goal of this issue is to revisit the [js files] to check if there is some outdated content or space for improvement.

Your question reminds me of the #102 issue.

[Ansys]: https://ansys.com
[Sphinx]: https://www.sphinx-doc.org/en/master/
[css files]: https://github.com/ansys/ansys-sphinx-theme/tree/main/src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/css
[js files]: https://github.com/ansys/ansys-sphinx-theme/tree/main/src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/js

Revathyvenugopal162:
Moving this to V1.5 as i have some major refactoring from my side also considering #591 .

thecrusader25225:
Is this still open? can you assign me?

PR #798, opened 2025-09-15
feat: cleanup ``js``

Revathyvenugopal162:
Fix https://github.com/ansys/ansys-sphinx-theme/issues/532

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,27 +1,27 @@
 repos:
 
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.8.6
+  rev: v0.14.1
   hooks:
     - id: ruff
       args: [--fix]
     - id: ruff-format
 
 - repo: https://github.com/codespell-project/codespell
-  rev: v2.3.0
+  rev: v2.4.1
   hooks:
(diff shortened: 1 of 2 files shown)


src/ansys_sphinx_theme/theme/ansys_sphinx_theme/components/announcement_version.html
```diff
@@ -1,14 +1,28 @@
...
+    .then((text) => {
+      var msgDiv = document.getElementById("announcement_msg");
+      if (msgDiv) {
+        msgDiv.innerHTML = text;
+      }
+    })
+    .catch((error) => {
+      // Log only unexpected fetch errors
+      if (!(error.message && error.message.startsWith("HTTP error: 404"))) {
+        console.error(`Fetch problem: ${error.message}`);
```
jorgepiloto: Let's indicate a bit more where the error applies, just to easily filter it in the console output:

> Could net fetch announcement.html

jorgepiloto (changes requested):
Just left one minor comment about an error message.

Do you think we could add in this same pull request <https://djlint.com/> or an alternative? I am not sure if we are formatting our HTML template files.

PR #811, opened 2025-10-08
feat: adapt new logo

Revathyvenugopal162:
Add updated ansys logo


Revathyvenugopal162:
> If these are the new logos, let's just override the old images. Also, looks like you removed the PyAnsys logos.

Replaced the logos; the removed ones were unused and not imported in the theme.

PR #813, opened 2025-10-08
ci: add pull request preview action

Revathyvenugopal162:


RobPasMue (approved):
LGTM!

jorgepiloto (changes requested):
This action needs to add the conditional `if: github.event.action != 'closed'` in every other job to avoid re-running once the pull is closed.

RobPasMue (approved):
LGTM

On a side note.. Moving forward, could we embed this "if" behavior as part of the action itself? We could expose an input parameter for the "event-action" and skip things or run things depending on the status. What do you think?

PR #832, opened 2025-11-05
feat: add dataclass and pydantic class example

mhoeijm:


moe-ad (approved):
Thanks for the contribution @mhoeijm.

moe-ad:
Adding this comment for full context:
Martijn just synced with me offline and the goal of this PR was originally to demonstrate a limitation of `autoapi`, which is not picking up the description of a `Field` as an attribute's docstring (oddly what I "corrected" before merging).

@RobPasMue should we leave it merged or I should open a PR to revert the changes?
Or you could revert main to the previous commit and I can just reopen this PR.

PR #928, opened 2026-03-27
feat: add weighted score

### LINKED ISSUE #591: Use weights to sort results in the static search
It would be great if we could implement some sort of metric or "weights" to sort the results presented by the static search.

Revathyvenugopal162:
fix #591 
## Summary

Enhances the search functionality by introducing weighted scoring to improve result relevance. The search bar UI and filter display logic were also refined for better usability.

## Changes

- Added weighted score logic to search ranking
- Refactored search JavaScript (`search.js`, `search-main.js`) for improved behavior and filter display
- Updated search styles in `ast-search.scss`
- Minor updates to search backend (`fuse_search.py`, `search/__init__.py`)
- Updated `search.html` template
  - Refactored keyboard navigation and result rendering logic., `enter` will take to the `search` page


src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/js/search.js
jorgepiloto: Please remove this text. The shortcut changed and we now use only `Enter` when no default suggestion is selected.

src/ansys_sphinx_theme/search/__init__.py
```diff
@@ -32,9 +32,16 @@ def update_search_config(app: Sphinx) -> None:
...
+        {"name": "section", "weight": 3},
+        {"name": "title", "weight": 3},
+        {"name": "text", "weight": 1},
+        {"name": "objectID", "weight": 0.5},
+    ]
     theme_static_options["threshold"] = theme_static_options.get("threshold", 0.2)
     theme_static_options["limit"] = theme_static_options.get("limit", 10)
+    theme_static_options["includeScore"] = True
+    theme_static_options["includeMatches"] = True
     app.config.html_theme_options["static_search"] = theme_static_options
```
RobPasMue, line 62: These weights and options should be configurable IMO. And here you are simply overriding any value passed in by the user. I think we should document them here - https://sphinxdocs.ansys.com/version/dev/user-guide/options.html#static-search

RobPasMue: And of course, change the logic to include defaults - and not simply override

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/js/search-main.js
```diff
@@ -104,10 +97,14 @@ require(["fuse"], function (Fuse) {
   };
 
   /**
-   * Initialize the search system by loading the document search index and library indexes.
-   * Sets up filter dropdowns and triggers initial search if needed.
+   * Load the main search index and all library indexes, then set up the filter UI.
    */
   async function initializeSearch() {
```
jorgepiloto, line 102: Nice fix. One more guard will make this path safe: return early in `showLibraryDropdown()` when `#library-dropdown` is missing.

```js
function showLibraryDropdown() {
  const dropdown = document.getElementById("library-dropdown");
  if (!dropdown) return;
  // ...
}
```

That avoids the null dereference and keeps behavior the same when `search_extra_sources` is configured.

jorgepiloto (changes requested):

src/ansys_sphinx_theme/search/__init__.py
Revathyvenugopal162: sure updated the docs, and add default values, thank you

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/js/search.js
Revathyvenugopal162: removed the session

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/js/search-main.js
Revathyvenugopal162: thanks, added this

src/ansys_sphinx_theme/search/__init__.py
RobPasMue: Thanks @Revathyvenugopal162

doc/source/user-guide/options.rst
```diff
@@ -156,27 +156,61 @@ To exclude files or directories from the search index, you can use the
 of strings representing the directories or files to exclude from the search
 index.
 
-Here is an example of how to add the ``static_search`` dictionary to the
-``html_theme_options`` dictionary:
+Search result ranking
+~~~~~~~~~~~~~~~~~~~~~
+
+By default, the theme assigns the following Fuse.js field weights so that
+section headings and page titles rank above body text:
```
RobPasMue:
```suggestion
By default, the theme assigns the following Fuse.js field weights so that
section headings and page titles rank over the body text:
```

RobPasMue (approved):
LGTM - minor comment

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/js/search-main.js
```diff
@@ -119,33 +118,45 @@ require(["fuse"], function (Fuse) {
       searchData = data;
       fuse = new Fuse(searchData, SEARCH_OPTIONS);
 
+      // Render filter UI immediately after the main index is ready.
+      // External library fetch failures should not block sidebar rendering.
+      showObjectIdDropdown();
```
jorgepiloto, line 123: Good catch handling missing `#search-sidebar` in `setupFilterDropdown()`, but `initializeSearch()` still calls `showObjectIdDropdown()` unconditionally. If the sidebar is missing, `showObjectIdDropdown()` hits `null` at `dropdown.innerHTML` and initialization stops. Please guard this call (or return early in `showObjectIdDropdown()`) so search still works on pages without sidebar filters.

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/js/search-main.js
```diff
@@ -299,33 +311,63 @@ require(["fuse"], function (Fuse) {
...
   /**
-   * Perform the search and update the results UI.
+   * Compute a relevance score for a Fuse.js result.
+   * Fuse score: 0 = perfect match, 1 = complete mismatch; boosted by matched field weight.
+   * @param {Object} result - Fuse.js result with .score and .matches.
+   * @returns {number}
+   */
+  function computeRelevance(result) {
+    const fieldWeights = { section: 3, title: 2, text: 1, objectID: 0.5 };
+    const matches = result.matches || [];
```
jorgepiloto, line 328: The new re-ranking hardcodes `title` weight to `2`, while `static_search` defaults/docs use `title = 3`. That can produce ordering that differs from configured and documented behavior. Please read weights from `SEARCH_OPTIONS.keys` (or align these constants with the defaults) so ranking stays consistent.

jorgepiloto (changes requested):
Thanks for the weighted-search work and the UI cleanup. I found two blocking issues that should be fixed before merge.

1. **request** (`src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/js/search-main.js`, around `showObjectIdDropdown()` call in `initializeSearch`): `setupFilterDropdown()` now supports layouts without `#search-sidebar`, but `initializeSearch()` still unconditionally calls `showObjectIdDropdown()`. In no-sidebar layouts this dereferences `null` and aborts initialization.

2. **request** (`src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/js/search-main.js`, `computeRelevance`): re-ranking hardcodes `title: 2`, but defaults/docs set `title` weight to `3` via `static_search.keys`. This makes ranking behavior inconsistent across backend config, docs, and runtime.

Backward compatibility note: the PR changes ranking semantics; keeping JS re-ranking aligned with `static_search.keys` is important so existing user configuration behaves predictably.

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/js/search-main.js
Revathyvenugopal162: you are right, let me fix this

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/js/search-main.js
Revathyvenugopal162: i will add an `if`with `!not(dropdown)`, good catch, thanks

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/js/search-main.js
Revathyvenugopal162: addressed in [272b991](https://github.com/ansys/ansys-sphinx-theme/pull/928/commits/272b991421dbf88c5c4f5124673b29b58283ff69), thank you @jorgepiloto

jorgepiloto (approved):
Thanks for the follow-up updates. I rechecked the two blocking points from my last review and both are now fixed.

- `showObjectIdDropdown()` now safely returns when `#objectid-dropdown` is absent, so pages without the sidebar no longer break search initialization.
- Re-ranking now derives field weights from `SEARCH_OPTIONS.keys` (with aligned fallback defaults), so runtime ranking stays consistent with the configured/documented `static_search.keys` behavior.

Approving.

Revathyvenugopal162:
thank you @jorgepiloto  and @RobPasMue  for the review and valuable suggestions. Merging this PR

PR #955, opened 2026-04-24
fix: `backstop` tests issue with collapsible sidebar

Revathyvenugopal162:
part of #932 

BackstopJS by default captures the full scrollable page height. For long documentation pages, this caused the position: sticky sidebar to stretch to the full document height, producing screenshots that looked nothing like what a real browser renders. Using "selectors": ["viewport"] restricts capture to the visible window area, matching real browser behaviour.


RobPasMue (approved):
LGTM! Thanks for fixing this @Revathyvenugopal162

PR #943, opened 2026-04-13
feat: Add  option to globally hide the secondary sidebar page TOC

Revathyvenugopal162:
fixes #703 

# Secondary component in primary sidebar 

If the theme option, enabled, you can add the seconday sidebar on this page in primary 
Screenshot 2026-04-20 at 16 43 13

Also initial request , Making the secondary sidebar bold and aligned with Design webpage
Screenshot 2026-04-20 at 17 20 15

**New sidebar TOC and source button options:**

* Added support for `show_page_toc`, `show_page_toc_in_primary_sidebar`, and `show_source_button` options in `html_theme_options`, allowing users to hide the page TOC globally, move it into the primary sidebar, or hide the "View source" button. The theme logic now processes these options and updates the sidebar layout accordingly. 
* Updated the primary sidebar template (`sidebar-nav-bs.html`) to inject the page TOC as a collapsible, scroll-spy-enabled list under the current page entry when `show_page_toc_in_primary_sidebar` is enabled.

**Documentation updates:**

* Expanded the user guide (`options.rst`) with detailed instructions and code examples for controlling the secondary sidebar, moving the TOC, hiding the "View source" button, and customizing the TOC depth, both globally and per page.

**Styling and layout improvements:**

* Updated SCSS styles to support the new sidebar TOC behaviors, including hiding the primary sidebar when empty, adjusting widths based on sidebar presence, and styling the injected TOC and toggle button.

PR #983, opened 2026-05-21
fix: cheatsheet configurations in sidebar

Revathyvenugopal162:
Cheatsheet was not visible in the sidebar with latest release, because we were hiding all the components in the sidebar. 

After #943 

The two other bugs (wrong Jinja variable, missing context variable)


Revathyvenugopal162:
testing this in https://github.com/ansys/pyspeos/pull/973

AlejandroFernandezLuces (approved):
LGTM!

Issue #998, opened 2026-05-28
Copyright footer could be not updated


carlosegomezc:
### 🔍 Before submitting the issue

- [x] I have searched among the existing issues
- [x] I am using a Python virtual environment

### 🐞 Description of the bug

According to the communication given to the R&D teams, the new copyright should be:

`Copyright © <YEAR> Synopsys, Inc. and ANSYS, Inc. All rights reserved.`

Is it in your scope applies this change?

### 📝 Steps to reproduce

Use the last version of the theme.

### 💻 Which operating system are you using?

Windows

### 🐍 Which Python version are you using?

3.13

### 📦 Installed packages

```shell
accessible-pygments==0.0.5
alabaster==1.0.0
-e git+https://ado.internal.synopsys.com/tfs/ANSYS_Development/SBU-SCADE-s-one/_git/ANSYS.SONE.API.pyscadeone@ac0e8087486884bb220ed34e56b037cade2bfe67#egg=ansys_scadeone_core
ansys-sphinx-theme==1.7.2
anyio==4.13.0
attrs==26.1.0
babel==2.18.0
beautifulsoup4==4.14.3
certifi==2026.5.20
cffi==2.0.0
cfgv==3.5.0
charset-normalizer==3.4.7
click==8.4.1
clr_loader==0.2.10
cmake==4.3.2
codespell==2.4.2
colorama==0.4.6
coverage==7.14.1
distlib==0.4.0
docutils==0.22.4
fastjsonschema==2.21.2
filelock==3.29.0
flit==3.12.0
flit_core==3.12.0
FMPy==0.3.29
h11==0.16.0
identify==2.6.19
idna==3.16
imagesize==2.0.0
importlib_metadata==9.0.0
iniconfig==2.3.0
Jinja2==3.1.6
jsonschema==4.26.0
jsonschema-specifications==2025.9.1
jupyter_core==5.9.1
lark==1.3.1
lxml==6.1.1
MarkupSafe==3.0.3
msgpack==1.1.2
nbformat==5.10.4
nodeenv==1.10.0
numpy==2.4.6
numpydoc==1.10.0
packaging==26.2
pdf2image==1.17.0
pillow==12.2.0
platformdirs==4.9.6
pluggy==1.6.0
pre_commit==4.6.0
pycparser==3.0
pydata-sphinx-theme==0.16.1
Pygments==2.20.0
pyrefly==0.63.1
pytest==9.0.3
pytest-cov==7.1.0
[...]

PR #1004, opened 2026-06-08
feat: add default `copyright` options with theme

Revathyvenugopal162:
as part of #998

### DIFF
--- a/doc/source/user-guide/configuration.rst
+++ b/doc/source/user-guide/configuration.rst
@@ -11,10 +11,30 @@ To use the Ansys Sphinx Theme, add the following line to your project's Sphinx `
 
 The Ansys Sphinx Theme provides these features:
 
+- `Copyright`_
 - `PyAnsys and Ansys logos`_
 - `Version switcher`_
 - `PDF cover page`_
 - `Custom CSS`_
+
+Copyright
+~~~~~~~~~
+
+The Ansys Sphinx Theme includes a copyright notice in the footer of the documentation.
+To set the copyright notice, add the following code to your project's Sphinx ``conf.py`` file:
+
+.. code-block:: python
+
+   copyright = "Copyright © <YEAR> Synopsys, Inc. and ANSYS, Inc. All rights reserved."
+
+customise the year dynamically, you can use the following code:
+
+.. code-block:: python
+
+   from datetime import datetime
+
+   current_year = datetime.now().year
+   copyright = f"Copyright © {current_year} Synopsys, Inc. and ANSYS, Inc. All rights reserved."
 
 PyAnsys and Ansys logos
 ~~~~~~~~~~~~~~~~~~~~~~~
--- a/doc/source/user-guide/configuration.rst
+++ b/doc/source/user-guide/configuration.rst
@@ -21,13 +21,15 @@ Copyright
 ~~~~~~~~~
 
 The Ansys Sphinx Theme includes a copyright notice in the footer of the documentation.
-To set the copyright notice, add the following code to your project's Sphinx ``conf.py`` file:
+To define a copyright notice, add the following to your project’s ``conf.py`` file:
 
 .. code-block:: python
 
    copyright = "Copyright © <YEAR> Synopsys, Inc. and ANSYS, Inc. All rights reserved."
 
-customise the year dynamically, you can use the following code:
+Replace <YEAR> with the appropriate year.
+
+To automatically update the year, use the following:
 
 .. code-block:: python
 
@@ -35,6 +37,8 @@ customise the year dynamically, you can use the following code:
 
    current_year = datetime.now().year
    copyright = f"Copyright © {current_year} Synopsys, Inc. and ANSYS, Inc. All rights reserved."
+
+This ensures the copyright year remains up to date without manual changes.
 
 PyAnsys and Ansys logos
 ~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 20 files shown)


doc/source/user-guide/configuration.rst
```diff
@@ -11,11 +11,35 @@ To use the Ansys Sphinx Theme, add the following line to your project's Sphinx `
...
+~~~~~~~~~
+
+The Ansys Sphinx Theme includes a copyright notice in the footer of the documentation.
+To define a copyright notice, add the following to your project’s ``conf.py`` file:
+
+.. code-block:: python
+
+   copyright = "Copyright © <YEAR> Synopsys, Inc. and ANSYS, Inc. All rights reserved."
+
+Replace <YEAR> with the appropriate year.
```
klmcadams:
```suggestion
Replace <YEAR> with the appropriate year or year range.
```

doc/source/user-guide/configuration.rst
```diff
@@ -11,11 +11,35 @@ To use the Ansys Sphinx Theme, add the following line to your project's Sphinx `
...
+
+To automatically update the year, use the following:
+
+.. code-block:: python
+
+   from datetime import datetime
+
+   current_year = datetime.now().year
```
klmcadams: Should there be another line to retrieve the start year of the repository similar to the logic in the license headers pre-commit hook?

doc/source/user-guide/configuration.rst
```diff
@@ -11,11 +11,35 @@ To use the Ansys Sphinx Theme, add the following line to your project's Sphinx `
...
 
+Copyright
+~~~~~~~~~
+
+The Ansys Sphinx Theme includes a copyright notice in the footer of the documentation.
+To define a copyright notice, add the following to your project’s ``conf.py`` file:
+
+.. code-block:: python
+
+   copyright = "Copyright © <YEAR> Synopsys, Inc. and ANSYS, Inc. All rights reserved."
```
klmcadams: @RobPasMue Should I update the default copyright in the add-license-headers script (https://github.com/ansys/pre-commit-hooks/blob/c5198e0c211720e34dd097ae57dc8153eda8e269/src/ansys/pre_commit_hooks/add_license_headers.py#L42) to match this line?

klmcadams (commented):
Looks good to me! Just had some questions in relation to the pre-commit hook repo

doc/source/user-guide/configuration.rst
RobPasMue: This has been discussed several times with @MaxJPRey offline... I believe that until we are not requested to do it, we should not modify it.

If we got final confirmation from the legal team, then fine by me.

RobPasMue (commented):
Same comment as below.. we really need to get confirmation on this before we do the change.

Revathyvenugopal162:
@RobPasMue and @klmcadams,
Please note that I haven’t changed the copyright in this documentation. I have only added guidance for users on how to update the copyright themselves, so that we can provide this as a reference link.

Also Maxime confirmed that this copyright can be used. But from what i understood it is not mandatory for us at the moment. That’s why I did not modify the copyright in the documentation itself. Instead, I used the new version only as an example in the documentation.

doc/source/user-guide/configuration.rst
Revathyvenugopal162: Usually in docs footer, it is just year, not year range

RobPasMue:
> @RobPasMue and @klmcadams, Please note that I haven’t changed the copyright in this documentation. I have only added guidance for users on how to update the copyright themselves, so that we can provide this as a reference link.

Right - I understand that. However, I hope you agree that guidance is also a way to encourage people to do it... and misalign.

My point is that we should align all our libraries... I don't think it's a good idea that some people start defining their own Ansys & Synopsys headers on their own... because when they do it, they no longer follow the defaults we ship.. which will lead to stale content in the long run. I am more in favor of exposing the new license header - rather than encouraging/opening-the-door to people for modifying it themselves

MaxJPRey (approved):

RobPasMue (approved):
Got things ready in https://github.com/ansys/pre-commit-hooks/pull/445 as well

src/ansys_sphinx_theme/__init__.py
```diff
@@ -601,6 +602,30 @@ def replacer(match):
...
+    Parameters
+    ----------
+    app : Sphinx
+        Sphinx application instance for rendering the documentation.
+
+    Notes
+    -----
+    This function checks if the 'copyright' configuration is already set by the user.
+    If it is set, it logs a message indicating that the user's configuration
+    will overwrite the default
```
SMoraisAnsys:
```suggestion
    will overwrite the default.
```

src/ansys_sphinx_theme/__init__.py
```diff
@@ -601,6 +602,30 @@ def replacer(match):
...
+    """
+    copyright = getattr(app.config, "copyright", None)
+    if copyright:
+        logging.getLogger(__name__).info(
+            "The 'copyright' configuration adding from ansys-sphinx-theme is being overwritten by the user's configuration. "  # noqa: E501
```
SMoraisAnsys:
```suggestion
            "The user's 'copyright' configuration is being overwritten by ansys-sphinx-theme's default."  # noqa: E501
```
I'm a bit confused by the message and my understanding of the actual behavior. Isn't that changing the project's copyright configuration ? If yes, this message might be clearer.

PR #1023, opened 2026-06-22
feat: announcements with theme

### LINKED ISSUE #1043: Announcement banner theme & styling enhancement
Enhance the announcement banner UI while keeping the existing logic unchanged.

### Changes Required

- Apply theme-based styling to announcement banners.
- Restore the old banner design/style without changing functionality.
- Add announcement customization options through theme settings.
- Apply different styles based on announcement type:
   - Info
   - Success
   - Warning
  - Error

- Add a button to open announcement details in a popup/modal.
- Use session storage to prevent dismissed banners from reappearing during the same user session.

### Things should consider

- Existing banner logic remains unchanged.
- Banner appearance is configurable through theme settings.
- Different announcement types have distinct styling.
- Popup button is available and works correctly.
- Dismissed banners do not reappear until a new session starts.

Revathyvenugopal162:
Fix #1043 
Adds a new fully customisable announcement banner component to the theme,
replacing the previous minimal version-only banner.

### Feature
- **New `announcement_banner` option** — accepts a single dict or a list of dicts
  with `type` (`info` / `warning` / `error` / `success`), `message`, and `url` fields.
- **Multi-notification modal** — when multiple announcements are configured, a
  "View" button opens a modal listing all notifications with per-type icons.
- **Legacy `announcement.html` support** — existing `announcement.html` files are
  still fetched and merged into the modal alongside the new config-based items.
- **Per-type styling** — banner and modal border/icon colour follow the admonition
  CSS variables; background uses a subtle `color-mix` tint.

### UX
- **Dismiss persistence** — close button stores the dismissed state in
  `LocalStorage`, keyed by a SHA-1 hash of the banner content (djb2 fallback
  for plain HTTP). Because the key is content-based rather than URL-based, a new
  or updated announcement always reappears — even on patch releases that share the
  same docs URL.
- **Layout alignment** — banner width matches `.bd-content` (the middle column)
  and is centred with `margin: auto`, adapting correctly when the left sidebar
  is open or closed. Width is set to 80 % of the content column.

### Implementation
- Extracted all JS into `static/js/announcement_banner.js`; Jinja's `pathto()`
  base URL is injected via a `data-base-url` attribute on the `<script>` tag.
- Styles extracted into `assets/styles/announcement.scss`.
- Docs updated with full usage examples and configuration options in https://sphinxdocs.ansys.com/pull/1023/user-guide/options.html#announcement-banner


src/ansys_sphinx_theme/theme/ansys_sphinx_theme/components/ast_announcement.html
```diff
@@ -0,0 +1,492 @@
...
+                    btnEl.rel = "noopener noreferrer";
+                    btnEl.removeAttribute("data-modal");
+                    btnEl.style.display = "";
+                } else {
+                    btnEl.style.display = "none";
+                }
+            } else {
+                // Multiple notifications: update banner text + show popup button
+                if (textEl) {
+                    let summary = "There are " + total + " notification" + (total !== 1 ? "s" : "");
```
SMoraisAnsys:
```suggestion
                    let summary = "There are " + total + " announcement" + (total !== 1 ? "s" : "");
```

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/components/ast_announcement.html
```diff
@@ -0,0 +1,492 @@
...
+        // Parse ast items from JSON data attribute
+        let astItems = [];
+        try { astItems = JSON.parse(modalBody.dataset.astItems || "[]"); } catch(e) {}
+
+        const section = document.createElement("div");
+        section.classList.add("ast-announcement-extra");
+
+        const heading = document.createElement("div");
+        heading.classList.add("ast-notification-heading");
+        heading.textContent = "Notifications";
```
SMoraisAnsys:
```suggestion
        heading.textContent = "Announcements";
```

SMoraisAnsys:
Would it be possible to have a mixed approach where one 'main' announcement is kept above and can be specified by maintainers while the others would be packed below (in a second banner or below the first announcement).

mia-guo-ux:
@SMoraisAnsys @Revathyvenugopal162 
Thank you for the feedback. However, this mixed layout is not currently supported by the AWC banner component, which is designed to display announcements within a **single** banner structure. If we choose to customize the banner specifically for PyAnsys, we should also consider the visual impact: stacking multiple banners would likely make the header feel bulky and take up a significant amount of vertical space, as shown below:
image
image

Another option is to add a button that allows users to advance to the next announcement. However, please note that this interaction would not be bi-directional; the button would only support moving forward to the next announcement, rather than navigating back and forth between announcements.
image
image

SMoraisAnsys:
> @SMoraisAnsys @Revathyvenugopal162 Thank you for the feedback. However, this mixed layout is not currently supported by the AWC banner component, which is designed to display announcements within a **single** banner structure. If we choose to customize the banner specifically for PyAnsys, we should also consider the visual impact: stacking multiple banners would likely make the header feel bulky and take up a significant amount of vertical space, as shown below: image image
> 
> Another option is to add a button that allows users to advance to the next announcement. However, please note that this interaction would not be bi-directional; the button would only support moving forward to the next announcement, rather than navigating back and forth between announcements. image image

If we keep a single banner then I think that it would be best to see all announcements at once. When different level of announcement are mixed, I think that it would be good for the user to see above what we want to highlight (based on maintainer opinion) and keep the ability to identify critical information through colors. For example, I could highlight an information by putting it at the beginning of the stacked announcements and the user would still be able to identify quickly what is the "warning" / "critical" information.

mia-guo-ux:
@SMoraisAnsys @Revathyvenugopal162 

We should keep the banner **lightweight** because this is a technical documentation page, not a marketing page. Users are likely arriving with a task in mind. A heavy banner could reduce scannability, take attention away from the key documentation pathways, and push important content below the fold. The banner should establish context and confidence, while the body of the page should remain focused on fast access to docs, examples, and resources.

I’d suggest treating this as a **single summary banner**, not multiple visible announcement rows (like I proposed before). The banner can summarize the active announcements, for example: “There are 3 notifications that should be viewed: 2 urgent, 1 informational,” and use the highest active severity for the visual style. Users can then click **View** to see the full list.
image
image

A few design systems support avoiding stacked banners. Spectrum is explicit: “Don’t stack alert banners” and recommends not showing multiple alert banners at the same time. Carbon similarly recommends showing only one banner at a time for system/product-level messages. USWDS also warns that too many notifications can overwhelm or annoy users and are likely to be ignored. ([Adobe](https://spectrum.adobe.com/page/alert-banner/), [Spectrum](https://carbondesignsystem.com/patterns/notification-pattern/), [USWDS](https://designsystem.digital.gov/components/alert/))


Banners are innately color-coded according to the type of info they convey. 
image

Revathyvenugopal162:
> @SMoraisAnsys @Revathyvenugopal162
> 
> We should keep the banner **lightweight** because this is a technical documentation page, not a marketing page. Users are likely arriving with a task in mind. A heavy banner could reduce scannability, take attention away from the key documentation pathways, and push important content below the fold. The banner should establish context and confidence, while the body of the page should remain focused on fast access to docs, examples, and resources.
> 
> I’d suggest treating this as a **single summary banner**, not multiple visible announcement rows (like I proposed before). The banner can summarize the active announcements, for example: “There are 3 notifications that should be viewed: 2 urgent, 1 informational,” and use the highest active severity for the visual style. Users can then click **View** to see the full list. image image
> 
> A few design systems support avoiding stacked banners. Spectrum is explicit: “Don’t stack alert banners” and recommends not showing multiple alert banners at the same time. Carbon similarly recommends showing only one banner at a time for system/product-level messages. USWDS also warns that too many notifications can overwhelm or annoy users and are likely to be ignored. ([Adobe](https://spectrum.adobe.com/page/alert-banner/), [Spectrum](https://carbondesignsystem.com/patterns/notification-pattern/), [USWDS](https://designsystem.digital.gov/components/alert/))
> 
> Banners are innately color-coded according to the type of info they convey. image

Thank you Mia, i can confirm the current implementation is same as the initial design. please check the deployed docs.

Revathyvenugopal162:
i will add the docs for this feature now

doc/source/announcement.html
```diff
@@ -0,0 +1,24 @@
+<!-- djlint:off H021-->
+<!-- for inline styles -->
+<style>
```
Revathyvenugopal162: should remove before merging

doc/source/conf.py
```diff
@@ -79,6 +80,21 @@
...
+        {
+            "message": "This release is deprecated. Please upgrade.",
+            "type": "warning",
+            "link": "https://docs.pyansys.com/version/stable/",
+        },
+        {
+            "message": "Critical security patch required.",
+            "type": "error",
+        },
+    ],
```
Revathyvenugopal162, line 89:
```suggestion
```
cleanup the examples before merging

src/ansys_sphinx_theme/__init__.py
```diff
@@ -648,6 +648,10 @@ def setup(app: Sphinx) -> dict:
     # Add default HTML configuration
     setup_default_html_theme_options(app)
     load_navbar_configuration(app)
+    announcement_banner = app.config.html_theme_options.get("announcement_banner", {})
+    if announcement_banner:
+        print("Rendering announcement banner with the following configuration:")
+        print(announcement_banner)
```
Revathyvenugopal162:
```suggestion
```

doc/source/user-guide/options.rst
```diff
@@ -556,4 +587,47 @@ the ``conf.py`` file, add the following metadata at the very top of the file:
...
+
+.. code:: python
+
+    html_theme_options = {
+        "announcement_banner": [
+            {
+                "message": "Welcome to the Ansys Sphinx Theme documentation!",
+                "type": "info",
+            },
+            {
```
SMoraisAnsys, line 609: Could you add an example of success type here ?

SMoraisAnsys (approved):
LGTM. I checked the automation repo and I think the changes align. Once merged, released and used, could you monitor that everything is fine ?

doc/source/user-guide/options.rst
```diff
@@ -6,6 +6,27 @@ HTML theme options
...
+   :ref:`show_breadcrumbs`                      bool            Show breadcrumbs at the top of each page.
+   :ref:`add_hide_icons_in_navbar`              list            Add and hide icons in the navigation bar.
+   :ref:`static_search`                         list            Additional static search options to customize the search experience.
+   :ref:`cheat_sheets`                          list            Add a cheat sheet to the left navigation pane of your documentation.
+   :ref:`whats_new`                             list            Show a "What's new" section in the left navigation pane.
+   :ref:`secondary_sidebar`                     list            Control the secondary sidebar on each page.
+   :ref:`navigation_bar_dropdown`               list            Add a dropdown navigation bar to the top of your documentation.
+   :ref:`announcement_banner`                   list            Add an announcement banner to the top of your documentation pages.
+   ===========================================  ==============  ==========================================================================================
+
```
RobPasMue, line 27: Love it! But we are going to have to remember to keep it updated!

RobPasMue: Maybe add a pull request template with checkboxes... and say that if you include new options, make sure to update things here too

doc/source/conf.py
```diff
@@ -40,6 +40,7 @@
 html_theme = "ansys_sphinx_theme"
 html_short_title = html_title = "Ansys Sphinx Theme"
 html_static_path = ["_static"]
+html_extra_path = ["announcement.html"]
```
RobPasMue: You will also need to remove this one, right?

src/ansys_sphinx_theme/assets/styles/announcement.scss
RobPasMue, line 1: I would add some comment at the beginning of the file and potentially on each configuration explaining the purpose of it. Otherwise it might be difficult to identify later what each section is for (without using the dev tools on the browser)

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/components/announcement_version.html
```diff
@@ -1,22 +1,31 @@
 
 <!--
   Announcement Version Loader
-  Dynamically loads the announcement.html content into the announcement_msg div.
+  Dynamically loads the announcement.html content into:
+    1. #announcement_msg  — legacy banner bar (if present)
+    2. .ast-announcement-modal .ast-announcement-extra — new modal popup (if present)
 -->
-<div id="announcement_msg" aria-live="polite"></div>
+<!-- djlint:off H021-->
```
RobPasMue: I can see you are ignoring this rule again and again... doesn't it make sense to simply ignore it for the entire lib? If you really want to leverage inline styles you should be allowed to do so

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/js/announcement_banner.js
RobPasMue, line 1: Okay so... my two cents on this. This is an actual JS module. Same as if you had a Python module. Why don't we document functions properly like we do with Python? With all its parameters, return objects, proper description etc. Otherwise you need to look into the full method to understand what's going on

RobPasMue (commented):
Overall looks good! Thanks @Revathyvenugopal162 - left some enhancement suggestions!

doc/source/user-guide/options.rst
Revathyvenugopal162: Sure that is a great idea, i will add a PR template for theme including this in another PR if that works for you.

src/ansys_sphinx_theme/assets/styles/announcement.scss
Revathyvenugopal162: sure, i will do

doc/source/user-guide/options.rst
RobPasMue: Sounds good!

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/components/announcement_version.html
Revathyvenugopal162: The main reason I don't want to apply inline styles to all HTML files is that it's generally not considered a standard practice. However, there are some cases where it becomes unavoidable, especially for dynamic content such as announcements, search results, and similar components.

My intention was not to add this globally, but only in specific places where it's truly needed. If we add it to a global constant or apply it everywhere, there is  a risk that we may start using it in other files during testing and then forget to clean it up later during pushing the code.

Otherwise, I guess I just need to improve myself and avoid writing rough code 😄.

If everyone agrees with this approach, I can add it as a library rule/guideline so it's clear when and where it should be used

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/js/announcement_banner.js
Revathyvenugopal162: sure i will add those 2 docstrings for scss and js. thank you

doc/source/user-guide/options.rst
Revathyvenugopal162: sure . i can add

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/components/announcement_version.html
RobPasMue: Alright let's add it as a guideline in that case! Thanks @Revathyvenugopal162 for the explanation!

src/ansys_sphinx_theme/assets/styles/announcement.scss
Revathyvenugopal162: added the docstrings

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/js/announcement_banner.js
Revathyvenugopal162: added the js docstrings

doc/source/user-guide/options.rst
Revathyvenugopal162: done, thank you for pointing this

RobPasMue (approved):

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/js/announcement_banner.js
MaxJPRey, line 1: The PR description says the dismissed state is stored in `sessionStorage` (keyed by SHA‑1), but `applyDismiss` actually reads and writes `localStorage`. Either update the description or switch to `sessionStorage` to match the intended "reappear on new session" UX.
Right now a dismissal persists forever on that origin until the user clears storage.

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/js/announcement_banner.js
```diff
@@ -0,0 +1,376 @@
...
+  window.addEventListener("resize", alignToBdArticle);
+
+  const ATTENTION_TYPES = ["warning", "error"];
+
+  /**
+   * Function to detect the type of notification based on the background color of an element.
+   * @param {HTMLElement} el - The element to detect the type from
+   * @returns {string} - The detected type ("success", "warning", "info", etc.)
+   */
+  function detectType(el) {
```
MaxJPRey, line 104: `detectType()` creates a temporary `<span>`, appends it to `document.body`, reads computed style, then removes it, once per `<p>` in the file announcement.html.
This forces layout thrash for each paragraph.
I might be wrong but I think you can read `window.getComputedStyle(el).backgroundColor` directly on the original `<p>` (it's already in the DOM via the parser div) or on a detached element after setting the style
This would avoid mount/unmount.

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/js/announcement_banner.js
Revathyvenugopal162: great catch, i changed to local storage after we discuss in group. i will change the PR description, thanks

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/js/announcement_banner.js
Revathyvenugopal162: you are right, may be i can try directly fetching it from the element style  instead of creating a temporary span. also may be compute will not be required because the value should already be in RGB format. 

But i assume detached element + getComputedStyle) would NOT work. `getComputedStyle` on an element not in the live DOM returns "" for `backgroundColor` in most browsers, which is exactly why i tried to  attached to document.body. 

That said, it looks like we may be able to avoid all of that and just read the style directly. I'll give that a try.

src/ansys_sphinx_theme/assets/styles/announcement.scss
jorgepiloto, line 1: This file is missing the style for the `success` announcement type. Even if the style is compliant, I think it would be better to have its style next to the rest of the announcement types.

Revathyvenugopal162: Ah right, success was not in the initial implementation. i will add the style, good catch!

MaxJPRey (approved):

moe-ad (approved):
Late to the party, but can an image be added [here](https://sphinxdocs.ansys.com/pull/1023/user-guide/options.html#announcement-banner) to demonstrate the final result of the exact configuration used as the example? I think it will be a good visual guide.

A statement directing users to the "top of the current page to see how announcements typically render" can also be a good alternative.

Kudos for the great work @Revathyvenugopal162.

src/ansys_sphinx_theme/assets/styles/announcement.scss
```diff
@@ -0,0 +1,287 @@
...
+  align-items: center;
+  justify-content: center;
+
+  &.visible {
+    display: flex;
+  }
+}
+
+/* Announcement modal styles for the modal dialog that appears when "Learn more" is clicked. */
+.ast-announcement-modal {
```
jorgepiloto, line 163: This one is missing the `success` style.

src/ansys_sphinx_theme/assets/styles/announcement.scss
```diff
@@ -0,0 +1,287 @@
...
+  width: 1.4em;
+  height: 1.4em;
+  font: var(--fa-font-solid);
+  font-size: 1.1em;
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+}
+
+$announcement-types: (
```
jorgepiloto, line 62: This one is missing the `success` type.

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/js/announcement_banner.js
```diff
@@ -0,0 +1,372 @@
...
```
jorgepiloto, line 104: This function must be able to handle the `error` type.

doc/source/user-guide/options.rst
```diff
@@ -556,4 +587,51 @@ the ``conf.py`` file, add the following metadata at the very top of the file:
...
+- ``message``: Text displayed in the announcement banner.
+- ``type``: Type of announcement. Supported values are:
+
+  - ``info``: Informational messages.
+  - ``success``: Successful operations or positive updates.
+  - ``warning``: Warnings or deprecation notices.
+  - ``error``: Critical alerts or issues requiring immediate attention.
+
+  Announcement types are prioritized as follows: **error > warning > success > info**. If multiple announcements are configured, only the announcement with the highest priority is displayed.
```
jorgepiloto: I bet the current logic displays all the announcements, no matter if duplicated types exist. Can you confirm this, @Revathyvenugopal162?

jorgepiloto (commented):
Almost there, I left a couple of requests and one question about a difference between the documentation and the current logic.

doc/source/user-guide/options.rst
Revathyvenugopal162: yeah all announcement will display, only announcement with highest logic border color will be displayed as the main one's border. it was not clear. i will update that

src/ansys_sphinx_theme/assets/styles/announcement.scss
Revathyvenugopal162: done

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/js/announcement_banner.js
Revathyvenugopal162: @jorgepiloto This is the fn i am using to detect notifications from announcement.html and generate the announcement content. I intentionally did not include explicit error detection here because the "version is old" banner uses a color that is quite close to red. As a result, it could be classified as an error, causing the entire announcement banner to be displayed with error styling.

I do not have a strong objection to treating the "version is old" banner as an error, but I believe it is more appropriate to classify it as a warning since it is informational rather than indicating a system failure. That is why I removed the error logic from this function.

but i can add error detection if you would prefer the banner to be classified and displayed as an error.

Revathyvenugopal162:
> Late to the party, but can an image be added [here](https://sphinxdocs.ansys.com/pull/1023/user-guide/options.html#announcement-banner) to demonstrate the final result of the exact configuration used as the example? I think it will be a good visual guide.
> 
> A statement directing users to the "top of the current page to see how announcements typically render" can also be a good alternative.
> 
> Kudos for the great work @Revathyvenugopal162.

Right, i added 2 images. Thanks @moe-ad

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/static/js/announcement_banner.js
jorgepiloto: Thanks for the explanation. Makes sense.

jorgepiloto (approved):
Thanks for adding this. As we discussed, we'll try to make the theme version-aware in the near future so all this logic gets simplified both here and in the ansys/actions.

Revathyvenugopal162:
Thank you all for your review, suggestions and valuable commands. Merging this PR

PR #1101, opened 2026-08-25
fix: show actual section title in sidebar

### LINKED ISSUE #990: Give actual title instead of placeholder in sidebar
Image

Remove placeholder and add the actual title

jorgepiloto:
This replaces the primary sidebar placeholder with the actual section title. It resolves the title from the top breadcrumb when available, falls back to the current top-level sidebar item, updates the sidebar `aria-label`, and adds Playwright coverage for nested and top-navigation pages.

Fixes #990.

### DIFF
--- a/src/ansys_sphinx_theme/__init__.py
+++ b/src/ansys_sphinx_theme/__init__.py
@@ -451,8 +451,19 @@ def add_sidebar_context(
     doctree : docutils.nodes.document
         Document tree for the page.
     """
-    # Expose flag to Jinja templates (used by sidebar-nav-bs.html).
+    # Expose metadata to Jinja templates (used by sidebar-nav-bs.html).
     context["ast_page_toc_in_primary"] = getattr(app, "_ast_page_toc_in_primary", False)
+
+    print("Adding sidebar context for page:", pagename)
+    print("Title from context:", context.get("title"))
+    print("---------------------")
+    ast_section_title = context.get("title") or context.get("pagename") or "Section Navigation"
+    if not isinstance(ast_section_title, str):
+        ast_section_title = "Section Navigation"
+    ast_section_title = ast_section_title.replace("_", " ").strip()
+    if not ast_section_title:
+        ast_section_title = "Section Navigation"
+    context["ast_section_title"] = ast_section_title
 
     whatsnew_pages = whatsnew_sidebar_pages(app)
     cheatsheet_pages = cheatsheet_sidebar_pages(app)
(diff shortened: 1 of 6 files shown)


jorgepiloto (commented):
One non-blocking thing I've noticed is that `tests/test_sidebar_toc.spec.js` (the new `expectedTitle` evaluator block around lines ~80-145 in this PR) re-implements nearly the same selection logic as `sidebar-nav-bs.html`. That can let the test and implementation drift together and still pass. Consider asserting a concrete expected title for `NESTED_PAGE` (or deriving it through a simpler independent check) so the test protects behavior rather than mirroring internals.

jorgepiloto:
Thanks for your implementation, @Revathyvenugopal162. It is cleaner than the one I originally implemented.

jorgepiloto:
I just updated the tests and CI/CD is passing now.

src/ansys_sphinx_theme/__init__.py
```diff
@@ -428,6 +428,39 @@ def configure_theme_logo(app: Sphinx):
         theme_options["logo"] = logo_option
 
 
+def _normalize_sidebar_title(value: Any) -> str:
+    """Normalize sidebar title values from template context."""
+    if not isinstance(value, str):
+        return ""
+    return value.replace("_", " ").strip()
```
MaxJPRey: Few question for my understanding:

Underscores inside real titles will get mangled. 

If a page's actual title is `Using __init__.py` files, it becomes `Using   init  .py` files. Should underscore replacement only apply when the value is a filename/pagename fallback, not to human-authored titles?

Hyphens aren't handled. Sphinx pagenames like `getting-started` stay as-is, which is inconsistent with the underscore treatment. Is it expected? Just to be sure.

jorgepiloto: That's a good point. I'm checking this and can see some conflicts with the rendering of the API reference sections. Investigating this.

src/ansys_sphinx_theme/__init__.py
```diff
@@ -428,6 +428,39 @@ def configure_theme_logo(app: Sphinx):
         theme_options["logo"] = logo_option
 
 
+def _normalize_sidebar_title(value: Any) -> str:
+    """Normalize sidebar title values from template context."""
+    if not isinstance(value, str):
+        return ""
```
AlejandroFernandezLuces, line 453: Does this mean that the title would be empty?

AlejandroFernandezLuces (approved):
LGTM! Just left me thinking, is it worth setting up a small unit test module to capture cases like the one Maxime mentioned and make sure everything works?

PR #1105, opened 2026-08-28
feat: adding MCP Server banner feature

RobPasMue:
As title says. This will allow projects to add an admonition consistently across the ecosystem, with minimal configuration and allowing for MCP discoverability

image


moe-ad:
Why shouldn't projects just leverage the existing announcement banner component of the theme? Is it because we would like it to not be hidden behind the summary banner when there are multiple announcements?

RobPasMue:
> Why shouldn't projects just leverage the existing announcement banner component of the theme? Is it because we would like it to not be hidden behind the summary banner when there are multiple announcements?

There are several reasons for this indeed:

* Making sure that the MCP developments are not "missed out" by users: this is related to your point, we could leverage the announcements banner but in case you have multiple of them, users might oversee it. This is a request coming from PMs - to make it abundantly clear that the library has an MCP server available
* Standard messaging: although you didn't mention it... letting users define their own announcement implies that they will use different wordings for expressing the same. We want to avoid that by using the common message.

moe-ad:
> > Why shouldn't projects just leverage the existing announcement banner component of the theme? Is it because we would like it to not be hidden behind the summary banner when there are multiple announcements?
> 
> There are several reasons for this indeed:
> 
> * Making sure that the MCP developments are not "missed out" by users: this is related to your point, we could leverage the announcements banner but in case you have multiple of them, users might oversee it. This is a request coming from PMs - to make it abundantly clear that the library has an MCP server available
> * Standard messaging: although you didn't mention it... letting users define their own announcement implies that they will use different wordings for expressing the same. We want to avoid that by using the common message.

Got it... 👍

src/ansys_sphinx_theme/__init__.py
```diff
@@ -602,6 +602,54 @@ def replacer(match):
...
+
+    # Only inject context on the landing (index) page
+    index_page = app.config.root_doc or "index"
+    if pagename != index_page:
+        return
+
+    context["mcp_server"] = {
+        "url": url,
+        "project_name": mcp_options.get("project_name", "MCP Server"),
+    }
```
moe-ad, line 653: Since `mcp_server` being present in `html_theme_options` already signals the "intention" to use the option, shouldn't subsequent checks raise something like `ValueError` rather than just log a warning? That way, debugging config issues will be much clearer and easier.

I fear that in a situation where you get the config wrong in a project that already has a lot of doc-build warnings, this "silent" failure may be difficult to quickly pinpoint.

doc/source/user-guide/options.rst
```diff
@@ -691,3 +692,49 @@ When the user clicks on the ``view`` button, the banner expands to show all anno
...
+
+.. code-block:: python
+
+    html_theme_options = {
+        "github_url": "https://github.com/ansys/pymechanical",
+        "mcp_server": {
+            "url": "https://github.com/ansys/pymechanical-mcp",
+            "project_name": "PyMechanical MCP Server",
+        },
+    }
```
moe-ad, line 740:
```suggestion
```
The previous code block is enough IMO. This is a repetition.

moe-ad (commented):
Just some initial general comments. You also missed declaring the `mcp_server` option in [`src/ansys_sphinx_theme/theme/ansys_sphinx_theme/theme.conf`](https://github.com/ansys/ansys-sphinx-theme/blob/main/src/ansys_sphinx_theme/theme/ansys_sphinx_theme/theme.conf).

Revathyvenugopal162 (commented):
Few suggestions, 

- i think this should be implemented as a real Sphinx admonition rather than a custom banner, since it behaves like a note-style callout.
- Using an admonition would reduce the amount of custom HTML and CSS needed and would better match the theme’s existing admonition patterns.
- The feature should be treated as a supported theme option rather than a one-off template customization.
- Please register the theme option in the theme configuration so it is documented and available as part of the official theme API.

doc/source/conf.py
```diff
@@ -87,6 +87,10 @@
             "link": "https://sphinxdocs.ansys.com/version/dev/user-guide/options.html#announcement-banner",
         },
     ],
+    "mcp_server": {
+        "url": "https://github.com/ansys/pymechanical-mcp",
+        "project_name": "PyMechanical MCP Server",
+    },
```
RobPasMue: To be removed

src/ansys_sphinx_theme/theme/ansys_sphinx_theme/theme.conf
```diff
@@ -28,3 +28,4 @@ search_filters =
 navigation_dropdown =
 navbar_end = search-button-field, theme-switcher, navbar-icon-links
 announcement_banner =
+mcp_server =
```
RobPasMue, line 31: Thanks for adding it

Revathyvenugopal162:
Screenshot 2026-09-02 at 12 01 24
Output with admonitions

doc/source/conf.py
Revathyvenugopal162: removed

src/ansys_sphinx_theme/__init__.py
Revathyvenugopal162: done

doc/source/user-guide/options.rst
Revathyvenugopal162: We usually keep one example format and one realtime example in our theme options.  We have to cleanup all the theme options, so lets keep as it is now and cleanup the docs later

Revathyvenugopal162 (approved):

moe-ad (approved):

RobPasMue:
Can we get this merged and released @Revathyvenugopal162 ? =)

